PR #20, opened 2024-02-23
Skip proxy reads if collateral and protocol fee are zero

bbenligiray:
This allows us to not run data feeds when we set collateral and protocol fee to zero (which is the current state). I'll add the tests later.


Ashar2shahid (commented):
looks a little ugly but don't know if there is prettier way of doing it , maybe something like:

```
if (collateralInBasisPoints == 0 && protocolFeeInBasisPoints == 0){
 return (collateralAmount, protocolFeeAmount)
}

alikonuk1 (commented):
lgtm

PR #30, opened 2024-03-08
Make maximum subscription queue length customizable

bbenligiray:
I'm opening it as a draft PR because I didn't update the docs yet but feel free to check it out

### DIFF
--- a/contracts/api3-server-v1/Api3Market.sol
+++ b/contracts/api3-server-v1/Api3Market.sol
@@ -55,19 +55,6 @@ contract Api3Market is HashRegistry, ExtendedSelfMulticall, IApi3Market {
         bytes32 nextSubscriptionId;
     }
 
-    /// @notice Maximum subscription queue length for a dAPI
-    /// @dev Some functionality in this contract requires to iterate through
-    /// the entire subscription queue for a dAPI, and the queue length is
-    /// limited to prevent this process from being bloated. Considering that
-    /// each item in the subscription queue has unique update parameters, the
-    /// length of the subscription queue is also limited by the number of
-    /// unique update parameters offered in the dAPI pricing Merkle tree. For
-    /// reference, at the time this contract is implemented, the API3 Market
-    /// offers 4 update parameter options, and this number is not expected to
-    /// be increased (i.e., we do not expect this queue length limit to be hit
-    /// in practice).
-    uint256 public constant override MAXIMUM_SUBSCRIPTION_QUEUE_LENGTH = 5;
-
     /// @notice dAPI management Merkle root hash type
     /// @dev "Hash type" is what HashRegistry uses to address hashes used for
     /// different purposes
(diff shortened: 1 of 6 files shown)


contracts/api3-server-v1/Api3Market.sol
```diff
@@ -121,7 +119,17 @@ contract Api3Market is HashRegistry, ExtendedSelfMulticall, IApi3Market {
...
     /// @param proxyFactory_ ProxyFactory contract address
-    constructor(address owner_, address proxyFactory_) HashRegistry(owner_) {
+    /// @param maximumSubscriptionQueueLength_ Maximum subscription queue
+    /// length
+    constructor(
+        address owner_,
+        address proxyFactory_,
+        uint256 maximumSubscriptionQueueLength_
+    ) HashRegistry(owner_) {
+        require(
```
bbenligiray, line 129: Maybe also enforce 10 as the maximum value?

deploy/1_deploy.ts
```diff
@@ -11,6 +11,7 @@ import type { OwnableCallForwarder, ProxyFactory } from '../src/index';
 module.exports = async () => {
   const { deploy, log } = deployments;
   const [deployer] = await ethers.getSigners();
+  const MAXIMUM_SUBSCRIPTION_QUEUE_LENGTH = 10;
```
bbenligiray, line 14: !

contracts/api3-server-v1/Api3Market.sol
```diff
@@ -55,19 +55,6 @@ contract Api3Market is HashRegistry, ExtendedSelfMulticall, IApi3Market {
...
 
-    /// @notice Maximum subscription queue length for a dAPI
-    /// @dev Some functionality in this contract requires to iterate through
-    /// the entire subscription queue for a dAPI, and the queue length is
-    /// limited to prevent this process from being bloated. Considering that
-    /// each item in the subscription queue has unique update parameters, the
-    /// length of the subscription queue is also limited by the number of
-    /// unique update parameters offered in the dAPI pricing Merkle tree. For
-    /// reference, at the time this contract is implemented, the API3 Market
-    /// offers 4 update parameter options, and this number is not expected to
```
bbenligiray, line 66: We already want to do 5% (paired with OEV) so this is out the window

bbenligiray:
This PR will help me with https://github.com/api3dao/tasks/issues/626 too because I will run the exhaustive tests with a queue length of 3 (as the logic that works for 3 should also work for 5 or 10)

contracts/api3-server-v1/Api3Market.sol
acenolaza: I don't really see the value in locking ourselves (the only ones deploying this contract) with a fixed value other than us making a mistake. I would prefer to just add more details to the `@param` comment

acenolaza (approved):

contracts/api3-server-v1/Api3Market.sol
alikonuk1: I also agree with Santiago here.

alikonuk1 (approved):

contracts/api3-server-v1/Api3Market.sol
bbenligiray: Yeah I agree, which is why I omitted it in the first place

PR #34, opened 2024-03-19
Update Auctioneer registry and compute proxy address

Siegrift:
I've renamed the `metadata` to `data` to avoid confusion with proxy `metadata`. I used this naming convention in Auctioneer implementation as well. 

I also assume there is going to be a `bool` field specifying whether it's a dapi or dataFeed proxy. Maybe it could be `bytes32` in case we came up with more new proxies in the future. But it also seems unnecessary because the contract doesn't care about the schema and we can change it in case we need to.

Feel free to close, merge or edit. I can split it into multiple (more focused) PRs if you prefer.

### DIFF
--- a/contracts/api3-server-v1/AuctioneerRegistry.sol
+++ b/contracts/api3-server-v1/AuctioneerRegistry.sol
@@ -17,18 +17,17 @@ contract AuctioneerRegistry is Ownable, SelfMulticall {
 
     // This fee shouldn't guarantee indefinite support. We can specify that we
     // will deactivate proxies that haven't generated any OEV for a length of
-    // time.
-    // Similarly, we should be able to deactivate proxies whose metadata is not
-    // formatted correctly, or ones that are on chains that we're not serving
-    // yet, etc. Payment should not be reason to allow these to take up
+    // time. Similarly, we should be able to deactivate proxies whose data is
+    // not formatted correctly, or ones that are on chains that we're not
+    // serving yet, etc. Payment should not be reason to allow these to take up
     // auctioneer capacity.
     uint256 public activationFeeInUsdWith18Decimals;
 
     uint256 public maximumOevProxyCount;
 
-    mapping(bytes32 => bytes) public hashToOevProxyMetadata;
+    mapping(bytes32 => bytes) public hashToOevProxyData;
 
-    EnumerableSet.Bytes32Set private oevProxyMetadataHashSet;
+    EnumerableSet.Bytes32Set private oevProxyDataHashSet;
 
     constructor(
         address usdRateAirnode_,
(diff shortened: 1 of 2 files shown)


contracts/api3-server-v1/AuctioneerRegistry.sol
```diff
@@ -17,18 +17,17 @@ contract AuctioneerRegistry is Ownable, SelfMulticall {
...
+    // time. Similarly, we should be able to deactivate proxies whose data is
+    // not formatted correctly, or ones that are on chains that we're not
+    // serving yet, etc. Payment should not be reason to allow these to take up
     // auctioneer capacity.
     uint256 public activationFeeInUsdWith18Decimals;
 
     uint256 public maximumOevProxyCount;
 
-    mapping(bytes32 => bytes) public hashToOevProxyMetadata;
+    mapping(bytes32 => bytes) public hashToOevProxyData;
```
bbenligiray, line 28: I think I will call this "OEV proxy details" as in "data feed details" from AirseekerRegistry because they are quite similar

contracts/api3-server-v1/AuctioneerRegistry.sol
```diff
@@ -70,29 +69,30 @@ contract AuctioneerRegistry is Ownable, SelfMulticall {
...
     }
 
-    // The OEV proxy metadata is:
+    // The OEV proxy data is:
     //   - Chain ID
     //   - Data feed ID / dAPI name
+    //   - Is dAPI
     //   - Beneficiary address
-    //   - Metadata
+    //   - Data
```
bbenligiray, line 77: I'm referring to the actual/original proxy metadata here so it should remain so

src/proxy.ts
```diff
@@ -22,8 +22,18 @@ function getDeploymentAddresses(chainId: ethers.BigNumberish) {
...
+interface ContractAddresses {
+  api3ServerV1Address: string;
+  proxyFactoryAddress: string;
+}
+
+function computeDapiProxyAddress(
+  chainId: ethers.BigNumberish,
+  dapiName: string,
+  metadata: ethers.BytesLike,
+  contracts: ContractAddresses = getDeploymentAddresses(chainId)
```
bbenligiray, line 34: If one specifies `contracts`, `chainId` becomes redundant, so I'm not sure that I like this (or I misunderstood the syntax)

bbenligiray: I think one 
```
function computeDapiProxyAddress(
  chainId: ethers.BigNumberish,
  dapiName: string,
  metadata: ethers.BytesLike
  )
```
and one
```
function computeDapiProxyAddress(
  contracts: ContractAddresses,
  dapiName: string,
  metadata: ethers.BytesLike
  )
```
makes more sense but the second one is more error-prone and I don't see who would need it

bbenligiray:
I'll leave this open and address in https://github.com/api3dao/tasks/issues/912

contracts/api3-server-v1/AuctioneerRegistry.sol
```diff
@@ -70,29 +69,30 @@ contract AuctioneerRegistry is Ownable, SelfMulticall {
         // Emit event
     }
 
-    // The OEV proxy metadata is:
+    // The OEV proxy data is:
     //   - Chain ID
     //   - Data feed ID / dAPI name
+    //   - Is dAPI
```
bbenligiray, line 75: :+1:

bbenligiray:
> the contract doesn't care about the schema and we can change it in case we need to

This is the case but an Airnode ABI-like versioning (where the current version starts with an ASCII `1`) would be useful. Otherwise, it may be ambiguous to the auctioneer what version is used.

contracts/api3-server-v1/AuctioneerRegistry.sol
Siegrift: Yep, my bad.

src/proxy.ts
Siegrift: > If one specifies contracts, chainId becomes redundant, so I'm not sure that I like this (or I misunderstood the syntax)

You've understood it well. And, yeah the `chainId` is redundant in that case. I wanted my change to be backwards compatible with existing usage so it felt natural to me to make the "override" as a second argument.

> I think one ... and one ...

The problem with that is, you can't overload functions in JS, so this would end up as
```
function computeDapiProxyAddress(
  chainIdOrContracts: ethers.BigNumberish | ContractAddresses,
  dapiName: string,
  metadata: ethers.BytesLike
  )
```

which I don't like.

> I don't see who would need it

For example, we have e2e tests for auctioneer and use custom chains, for which there are no deployments in @api3/contracts. I'd be OK if we only had the second variant and called `getDeploymentAddresses(chainId)` in the auctioneer.

bbenligiray:
The updated Auctioneer v1 makes AuctioneerRegistry obsolete, so I'm closing this PR

Issue #39, opened 2024-03-23
Fix the race condition that causes unit tests to sometimes fail


bbenligiray:
See https://github.com/api3dao/contracts/actions/runs/8404100147/job/23015325687 for an example

PR #125, opened 2024-05-23
Replaces ethers.toBeArray() with ethers.getBytes() to fix the random …

acenolaza:
…Signature mismatch revert message in tests

Closes https://github.com/api3dao/contracts/issues/39

This was a nasty bug. The issue is basically using ethers.toBeArray() instead of ethers.getBytes() on a hex that starts with 0.  When this happens, toBeArray() drops the leading zeros so the signature will be different and the contract will revert when verifying them.

Here's a script to prove the fix works (needs to be run with hardhat):

```
import { ethers } from 'hardhat';

function main() {
    const hex =
        "0x005d11a5a4b257a8152ce171762c07a1279999dba84e2e55bf3aca46257eb0ee";

    const getBytes = ethers.getBytes(hex);
    console.log("🚀 ~ main ~ getBytes:", getBytes)

    const toBeArray = ethers.toBeArray(hex);
    console.log("🚀 ~ main ~ toBeArray:", toBeArray)

    // toBeArray internally calls this function
    // it basically takes the hex string and casts it to a bigint
    const toBigInt = ethers.toBigInt(hex);
    console.log("🚀 ~ main ~ toBigInt:", toBigInt)

    // then it converts it to back to a hex string
    const toString = toBigInt.toString(16)
    console.log("🚀 ~ main ~ toString:", toString)
    
    console.log("🚀 ~ main ~ original hex:", hex)
}

main();
```

### DIFF
--- a/deploy/2_update_beacon_set.ts
+++ b/deploy/2_update_beacon_set.ts
@@ -39,7 +39,7 @@ module.exports = async () => {
               ind + 1,
               `0x${(ind + 1).toString().padStart(64, '0')}`,
               await deployer?.signMessage(
-                ethers.toBeArray(
+                ethers.getBytes(
                   ethers.solidityPackedKeccak256(
                     ['bytes32', 'uint256', 'bytes'],
                     [templateId, ind + 1, `0x${(ind + 1).toString().padStart(64, '0')}`]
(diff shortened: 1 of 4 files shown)


bbenligiray (approved):
Thanks, great find. Those were originally `hexlify` in v5 and the [migration guide](https://docs.ethers.org/v6/migrating/) recommends that to be replaced with `toBeHex`, which behaves the same as `getBytes` in this context.

PR #126, opened 2024-05-24
DapiTransmuter

bbenligiray:
See https://github.com/api3dao/oev-searcher/issues/12#issuecomment-2129979004
L-M-K


test/utils/ExternalMulticallSimulator.sol.ts
```diff
@@ -0,0 +1,359 @@
...
+}
+
+async function getDapiTransmutationCalldata(
+  preTransmutationCalls: Call[],
+  postTransmutationCalls: Call[],
+  api3ServerV1Address: AddressLike,
+  api3ServerV1Interface: Interface,
+  dapiName: BytesLike,
+  value: BigNumberish,
+  externalMulticallSimulatorInterface: Interface
```
Siegrift, line 22: I find this a bit too complex. I'd separate this into two parts:
1. Function to create the call data for dAPI transmutation
2. The function that simulates the multicall and decodes the response.

See: https://github.com/api3dao/contracts/commit/284a823696f6f10458dbcafed628b84de5647fe7

test/utils/ExternalMulticallSimulator.sol.ts
```diff
@@ -0,0 +1,359 @@
...
+  // template ID would solve that.
+  const templateId = testUtils.generateRandomBytes32();
+  const timestamp = Math.floor(Date.now() / 1000);
+  const data = ethers.AbiCoder.defaultAbiCoder().encode(['int256'], [value]);
+  const signature = await testUtils.signData(airnode, templateId, timestamp, data);
+  const beaconId = testUtils.deriveBeaconId(airnode.address, templateId);
+  const transmutationCalls = [
+    {
+      target: api3ServerV1Address,
+      data: api3ServerV1Interface.encodeFunctionData('setDapiName', [dapiName, beaconId]),
```
Siegrift, line 37: Btw. this works great for dAPI proxies, but this won't work for data feed ones. I don't see how we could make that work though (so just thinking out loud).

test/utils/ExternalMulticallSimulator.sol.ts
```diff
@@ -0,0 +1,359 @@
...
+  const MOCK_AIRNODE_PRIVATE_KEY = '0x0fbcf3c01c9bcde58a6efa722b8d9019043dfaf5cdf557693442732e24b9f5ab';
+  const airnode = new ethers.BaseWallet(new ethers.SigningKey(MOCK_AIRNODE_PRIVATE_KEY));
+  // We want to use a Beacon ID that no one else has used to avoid griefing. Randomly generating the
+  // template ID would solve that.
+  const templateId = testUtils.generateRandomBytes32();
+  const timestamp = Math.floor(Date.now() / 1000);
+  const data = ethers.AbiCoder.defaultAbiCoder().encode(['int256'], [value]);
+  const signature = await testUtils.signData(airnode, templateId, timestamp, data);
+  const beaconId = testUtils.deriveBeaconId(airnode.address, templateId);
+  const transmutationCalls = [
```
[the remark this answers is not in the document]
bbenligiray: This is just an example. I didn't intend to do the actual SDK implementation.

test/utils/ExternalMulticallSimulator.sol.ts
bbenligiray: Since this is a DapiTransmuter, I left that out of scope. There's no way of making it work with data feeds anyway.

test/utils/ExternalMulticallSimulator.sol.ts
bbenligiray: I think making the provider call for the user is not great. I would have these functions instead
- `getDapiTransmutationCalls` (returns only two Calls for the transmutation)
- `encodeDapiTransmutationCalldata` (takes all Calls as argument, returns fully encoded calldata)
- The user is expected to use the calldata to make the call however they will
- `decodeReturnData` (takes returndata and all Calls as argument)

Call would include
- Target address
- [Human readable ABI](https://blog.ricmoo.com/human-readable-contract-abis-in-ethers-js-141902f4d917)
- Input args

which would allow us to decode the returndata

Siegrift: That works as well. We can separate the encoding & sending transaction, but imo we should also provide the function that makes the staticcall (accepts RPC provider and the calldata).

I like the `which would allow us to decode the returndata`.

contracts/utils/ExternalMulticallSimulator.sol
```diff
@@ -0,0 +1,34 @@
...
+/// @notice The contract only allows eth_call to be used for while making
+/// external calls to ensure that it is only used for simulating outcomes,
+/// rather than sending actual transactions
+contract ExternalMulticallSimulator is
+    SelfMulticall,
+    IExternalMulticallSimulator
+{
+    /// @dev Only allows calls where the sender is address-zero and the gas
+    /// price is zero, which indicates that the call is made through eth_call
+    modifier onlyEthCall() {
```
acenolaza, line 18: Nit: I might have called it `onlyVoidSignerEthCall()` instead

test/utils/ExternalMulticallSimulator.sol.ts
```diff
@@ -0,0 +1,359 @@
...
+        .multicall.staticCall(dapiTransmutationCalldata);
+      // Note that the decoding below is a bit tricky. From the array, the user picks
+      // the item whose returndata they want to read, decode that as a `bytes` type,
+      // then decode the result of that with the types of the returndata of the function
+      // they called.
+      const [transmutedValue] = ethers.AbiCoder.defaultAbiCoder().decode(
+        ['int224', 'uint32'],
+        ethers.AbiCoder.defaultAbiCoder().decode(['bytes'], returndata.at(-1)!)[0]
+      );
+      expect(transmutedValue).to.equal(transmutationValue);
```
acenolaza, line 356: This might help understand the double-decoding a bit better:

```suggestion
      const readReturndata = returndata.at(-1);
      const [decodedFunctionCallReturndata] = externalMulticallSimulator.interface.decodeFunctionResult(
        'functionCall',
        readReturndata!
      );
      const [transmutedValue] = ethUsdDapiProxyWithOev.interface.decodeFunctionResult(
        'read',
        decodedFunctionCallReturndata
      );
      expect(transmutedValue).to.equal(transmutationValue);
```

acenolaza (approved):

contracts/utils/ExternalMulticallSimulator.sol
bbenligiray: I think I should move the modifier in the function instead considering it's only used once, then we wouldn't need to name it too

bbenligiray: Done in https://github.com/api3dao/contracts/pull/126/commits/1962d79d46e463b9083d9b2b63b9de6779d48cd2

contracts/utils/ExternalMulticallSimulator.sol
```diff
@@ -0,0 +1,29 @@
...
+    /// @param target Target address of the external call
+    /// @param data Calldata of the external call
+    /// @return Returndata of the external call
+    function functionCall(
+        address target,
+        bytes memory data
+    ) external override returns (bytes memory) {
+        require(msg.sender == address(0), "Sender address not zero");
+        require(tx.gasprice == 0, "Tx gas price not zero");
+        return Address.functionCall(target, data);
```
alikonuk1, line 27: Can we also use Address.functionStaticCall here?

test/utils/ExternalMulticallSimulator.sol.ts
```diff
@@ -0,0 +1,364 @@
...
+  // Demonstrating that it's possible to call this contract with eth_sendTransaction,
+  // though it's not possible to use this to make external calls
+  describe('multicall via eth_sendTransaction', function () {
+    context('There is no call to be made', function () {
+      it('does not revert', async function () {
+        const { roles, externalMulticallSimulator } = await helpers.loadFixture(deploy);
+        await expect(externalMulticallSimulator.connect(roles.randomPerson).multicall([])).to.not.be.reverted;
+      });
+    });
+    context('The calls only include multicall()s with no calls no be made', function () {
```
alikonuk1, line 244:
```suggestion
    context('The calls only include multicall()s with no calls to be made', function () {
```

alikonuk1 (approved):
Doing this feels harder than how it can be done on a forked environment. This is an example of how I set any price on a forked environment; 
```
        vm.mockCall(
            INIT_ORACLE,
            abi.encodeWithSelector(IInitOracle(INIT_ORACLE).getPrice_e36.selector, undelying),
            abi.encode(10)
        );
```

bbenligiray:
> Doing this feels harder than how it can be done on a forked environment

This is to be expected but a forked chain is not performant enough to be considered an alternative in production use

PR #183, opened 2024-08-09
Refactor update beacon set mechanism

### LINKED ISSUE #182: Remove deployer restriction from `updateBeaconSet` phase
* Current configuration requires certain signer address to update test beacons 

**To-Do**
* Remove deployer address restriction

hiletmis:
Closes #182


bbenligiray (changes requested):
I'd like to keep the `signMessage()` flow as documentation. Check if the account is `0x07b589f06bD0A5324c4E2376d66d2F4F25921DE1`, use `signMessage()` if it is, use the hardcoded calldata otherwise.

hiletmis:
@bbenligiray including gas estimation signatures might not be a good idea. Removed it from script and updated the test call data for `nodary-rpc-watcher`

PR #189, opened 2024-08-13
Add hardware wallet support

hiletmis:
This PR introduces support for using a Ledger device as a signer.

To enable Ledger signing, users need to add an address to the environment file under the LEDGER_ACCOUNT tag. If this tag is not specified, the application will disregard any connected Ledger device and proceed with the default mnemonic deployer.

**There is a type conflict atm that should be solved prior to implementation**


bbenligiray:
As a note, this PR will be updated after the audit and @api3/chains is migrated into this repo

PR #248, opened 2024-11-18
Add dApps with IDs

bbenligiray:
I mainly need @Ashar2shahid to review the dApps. I tagged @hiletmis because we have been talking about moving chains over here.

The dApps need to be exported for the auctioneer config, DEAL UI, etc. and what is a better place to do that than contracts? It may bloat the package but we can worry about that later.


Ashar2shahid (approved):

hiletmis (approved):

bbenligiray:
I'll rework this

data/dapps/airpuff.json
```diff
@@ -0,0 +1,3 @@
+{
```
bbenligiray, line 1: I don't want to add a lot of fields to these and cause breaking changes later, only the alias is enough for now

src/index.ts
```diff
@@ -4,3 +4,4 @@ export { auctioneerMetadata };
 import deploymentAddresses from '../deployments/addresses.json';
 export { deploymentAddresses };
 export * from './proxy';
+export { DAPPS } from './generated/dapps';
```
bbenligiray, line 7: For example `data-feed-reader-example` needs to import this to prevent proxy deployments with unknown dApp aliases

src/proxy.ts
```diff
@@ -49,4 +49,30 @@ function computeApi3ReaderProxyV1Address(
...
+function computeDappSpecificApi3ReaderProxyV1Address(
+  dappAlias: string,
+  chainId: ethers.BigNumberish,
+  dapiName: string
+) {
+  return computeApi3ReaderProxyV1Address(chainId, dapiName, computeDappId(dappAlias, chainId), '0x');
+}
+
+export {
+  computeApi3ReaderProxyV1Address,
```
bbenligiray, line 74: Leaving this to not break the market, etc., and also for more flexible usage. The end users are only supposed to use `computeCommunalApi3ReaderProxyV1Address()` and `computeDappSpecificApi3ReaderProxyV1Address()` though.

data/dapps/airpuff.json
bbenligiray: We should at least add further identifying information in case when the aliases are no longer enough. That's nowhere near the case yet.

bbenligiray:
My thinking is that there are tiers of information about dApps. The DEAL UI and proxy deployment/address computation helpers merely need to know if a dApp exists, which this package does. The auctioneer needs to know on which chain each dApp is active. This will be detected automatically through DEAL NFT mints in the future, we need an interim solution for now (and that solution will not live in @api3/contracts because neither will the future solution). Finally, searchers need to know which data feeds the dApp uses on each chain, which will be handled by the searcher docs (so that will not be in @api3/contracts either).

bbenligiray:
After publishing this, I'll update https://github.com/api3dao/data-feed-reader-example and docs

src/proxy.ts
```diff
@@ -49,4 +49,30 @@ function computeApi3ReaderProxyV1Address(
...
+  return computeApi3ReaderProxyV1Address(chainId, dapiName, 1, '0x');
+}
+
+function computeDappId(dappAlias: string, chainId: ethers.BigNumberish) {
+  if (!DAPPS.some((dapp) => dapp.alias === dappAlias)) {
+    throw new Error(`dApp with alias ${dappAlias} not registered to the package`);
+  }
+  // Chain ID is a string to also support non-numerical (i.e., non-EVM) chain IDs.
+  // We use the BigNumberish type for now, until we need to support more arbitrary strings.
+  return ethers.solidityPackedKeccak256(['string', 'string'], [dappAlias, chainId.toString()]);
```
bbenligiray, line 62: ❗

src/proxy.ts
```diff
@@ -49,4 +49,30 @@ function computeApi3ReaderProxyV1Address(
...
 }
 
-export { computeApi3ReaderProxyV1Address };
+function computeCommunalApi3ReaderProxyV1Address(chainId: ethers.BigNumberish, dapiName: string) {
+  return computeApi3ReaderProxyV1Address(chainId, dapiName, 1, '0x');
+}
+
+function computeDappId(dappAlias: string, chainId: ethers.BigNumberish) {
+  if (!DAPPS.some((dapp) => dapp.alias === dappAlias)) {
+    throw new Error(`dApp with alias ${dappAlias} not registered to the package`);
```
bbenligiray, line 58: So the flow is the dApp applies to us to get a dApp ID assigned. We add the dApp to this package and publish, after which we deploy their proxies and let them verify the addresses. This friction is temporary, `dapps/` will be pre-populated with all known dApps out there once we switch to DEAL.

data/dapps/airpuff.json
Ashar2shahid: why would chainId be a breaking change ?

src/proxy.ts
```diff
@@ -49,4 +49,30 @@ function computeApi3ReaderProxyV1Address(
   );
 }
 
-export { computeApi3ReaderProxyV1Address };
+function computeCommunalApi3ReaderProxyV1Address(chainId: ethers.BigNumberish, dapiName: string) {
```
Ashar2shahid, line 52: "Communal" sounds weird, but I can't really think of any word except public, but that implies something is private which isn't necessarily true

data/dapps/airpuff.json
bbenligiray: I mean adding unneeded fields here and removing them later would be a breaking change so I won't add them until we're sure that we'll need them

Ashar2shahid (approved):

src/proxy.ts
bbenligiray: It's not a commonly used word so it works well as a term. It does have some negative connotations but we don't want people to use the communal proxy anyway.

data/dapps/airpuff.json
bbenligiray: Tracked in https://github.com/api3dao/contracts/issues/250

src/proxy.ts
```diff
@@ -49,4 +49,31 @@ function computeApi3ReaderProxyV1Address(
...
+  return computeApi3ReaderProxyV1Address(chainId, dapiName, 1, '0x');
+}
+
+function computeDappId(dappAlias: string, chainId: ethers.BigNumberish) {
+  if (!DAPPS.some((dapp) => dapp.alias === dappAlias)) {
+    throw new Error(`dApp with alias ${dappAlias} not registered to the package`);
+  }
+  return ethers.solidityPackedKeccak256(
+    ['bytes32', 'uint256'],
```
bbenligiray, line 61: I changed my mind on this. Chain ID can be a `uint256` and we can assign an arbitrary chain ID to non-EVM chains as necessary. Also, using the hash of the string rather than the string itself makes on-chain derivation of dApp IDs cheaper.

Issue #291, opened 2024-12-18
Validate deploymentAddresses


bbenligiray:
Sometimes `deploymentAddresses` can fall out of data because `deploy/3-document.ts` was not run

PR #303, opened 2024-12-26
Validate deployment addresses

bbenligiray:
Closes https://github.com/api3dao/contracts/issues/291

I felt like validate/verify didn't cut it to I came up with another verb, check.
- check: Checks if the local files containing metadata are correct before doing something that uses them, think of it like a flight check
- verify: Verifies bytecode
- validate: Validates the current on-chain state

As a note, previously when you added two chains to `data/chain-support.json` and ran the deployment script for the first chain, `3_document` crashed because it couldn't find the deployment files for the second chain, which would result in `4_verify` not running for the first chain. This meant that you had to loop through running the deployment script for the new chains twice. Now that `3_document` no longer exists, running the deployment script for the first chain will also attempt to verify it, which means a single loop should suffice.


hiletmis (approved):
LGTM. It was a thing and can be easily missed out. Thanks for the update.

PR #315, opened 2025-01-09
Derive proxy address over CLI

bbenligiray:
To test, clone the branch, then
```
pnpm pack
npx api3-contracts-13.3.0.tgz print-api3readerproxyv1-address --dapp-alias lendle --chain-id 59144 --dapi-name WBTC/USD
```


src/cli.ts
```diff
@@ -0,0 +1,72 @@
...
+      'chain-id': chainId,
+      'dapi-name': dapiName,
+    },
+    async (args) => {
+      const chain = CHAINS.find((chain) => chain.id === args['chain-id']);
+      if (!chain) {
+        throw new Error(`Chain with ID ${args['chain-id']} is not known`);
+      }
+      const marketUrl = `https://market.api3.org/${chain.alias}/${slugify(args['dapi-name'])}`;
+      if (await requestToUrlReturns404(marketUrl)) {
```
bbenligiray, line 60: Based on [this discussion](https://api3workspace.slack.com/archives/C03493F86Q7/p1736412086156099) this is not possible. I will replace this with an RPC call that checks the dAPI for a heartbeat update.

src/cli.ts
```diff
@@ -0,0 +1,72 @@
...
+// From https://github.com/api3dao/data-feeds/blob/main/packages/api3-market/src/utils/format.ts
+const slugify = (text: string) => text.toLowerCase().replaceAll(/[^\da-z-]+/g, '-');
+
+yargs(hideBin(process.argv))
+  .command(
+    'print-api3readerproxyv1-address',
+    'Prints the dApp-specific Api3ReaderProxyV1 address',
+    {
+      'dapp-alias': dappAlias,
+      'chain-id': chainId,
```
bbenligiray, line 51: This could also have been the chain alias. What I don't like about that is the user knows the chain ID but not necessarily the chain alias (and there is not a nice way of checking that), so this feels more user-friendly.

src/cli.ts
bbenligiray: Done in https://github.com/api3dao/contracts/pull/315/commits/f5de0d8e77649c9b3e282ee55b48d0a421823175

src/cli.ts
```diff
@@ -0,0 +1,96 @@
...
+        Api3ServerV1__factory.abi,
+        provider
+      ) as unknown as Api3ServerV1;
+      try {
+        const [, timestamp] = await api3ServerV1.readDataFeedWithDapiNameHash(
+          ethers.keccak256(ethers.encodeBytes32String(args['dapi-name']))
+        );
+        if (timestamp! + BigInt(24 * 60 * 60) < Date.now() / 1000) {
+          // eslint-disable-next-line no-console
+          console.log('⚠️ Feed timestamp appears to be stale');
```
bbenligiray, line 66: These are warnings considering that they may be caused by RPC issues

Ashar2shahid (approved):
looks good

PR #360, opened 2025-02-04
Stricter CLI command

bbenligiray:
The main change is
```
npx @api3/contracts print-api3readerproxyv1-address --dapp-alias lendle --chain-id 59144 --dapi-name WBTC/USD
```
now prints the below (and does not print the proxy address)
```
dApp alias: lendle
chain: Linea
dAPI name: WBTC/USD
⚠️ Proxy does not appear to have been deployed
```

To get the same behavior as before, you need to add a `--strict false` argument  as in
```
npx @api3/contracts print-api3readerproxyv1-address --dapp-alias lendle --chain-id 59144 --dapi-name WBTC/USD --strict false
```
which prints
```
dApp alias: lendle
chain: Linea
dAPI name: WBTC/USD
⚠️ Proxy does not appear to have been deployed
• Please confirm that https://market.api3.org/linea/wbtc-usd points to an active feed.
• Your proxy is at https://lineascan.build/address/0x32a5f22F630FAdcbD5C81E4cCfe6Bac98DC3c227
Please confirm that there is a contract deployed at this address before using it.
```

### DIFF
--- a/.changeset/heavy-lobsters-attend.md
+++ b/.changeset/heavy-lobsters-attend.md
@@ -1,0 +1,5 @@
+---
+'@api3/contracts': patch
+---
+
+Have the print-api3readerproxyv1-address CLI command print better looking errors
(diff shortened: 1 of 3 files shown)


bbenligiray:
This is released in 16.5.0

Issue #352, opened 2025-01-18
Merge api3dao/chains into this repo


bbenligiray:
api3dao/chains will be merged to this repo and @api3/contracts will include @api3/chains features, but @api3/chains will continue being maintained until all users migrate. One idea is to treat @api3/chains feature as beta versions until that happens by preceding them with `_` (which would mean that we can do breaking updates to them without a major release).

PR #395, opened 2025-02-18
Merge `@api3/chains` into `@api3/contracts`

bbenligiray:
Closes https://github.com/api3dao/contracts/issues/352

Pretty much a direct merge with some script naming changes, directory formatting and lint fixing. The most important change is that I omitted https://github.com/api3dao/chains/blob/main/scripts/ping-providers.ts assuming we don't depend on it for noticing dead providers, correct me if I'm wrong.

Also, the initial plan was to export `_CHAINS`, `_hardhatConfig()`, `_viemConfig()` and remove the underscore after they have settled, but looking at it again, that's probably unnecessary. So I used the original names with the intention being everyone migrating ASAP after release.

@andreogle for visibility


scripts/check-chains.ts
```diff
@@ -0,0 +1,62 @@
+import fs from 'node:fs';
```
bbenligiray, line 1: This was formerly called `validate-chains`. "Validate" refers to RPC calls to contracts in this repo, and "check" is better aligned with the lingo here.

src/hardhat-config.test.ts
```diff
@@ -0,0 +1,265 @@
...
+        accounts: { mnemonic: '' },
+        chainId: Number(chain.id),
+        url: `https://${chain.id}.xyz`,
+        ...overrides,
+      });
+    });
+  });
+
+  // eslint-disable-next-line jest/no-commented-out-tests
+  /*
```
bbenligiray, line 249: Not sure what else to do

src/utils/deep-equal.ts
```diff
@@ -0,0 +1,32 @@
+// This could be done with something like lodash's deepEqual, but in order
```
bbenligiray, line 1: This was previously under `src/validations/` but I don't see the point of having a separate directory just for this. Arguably these should be under `scripts/` and not `src/` anyway, but maybe not.

package.json
```diff
@@ -84,18 +89,20 @@
...
+    "ts-jest": "^29.2.5",
     "ts-node": "^10.9.2",
     "typechain": "^8.3.2",
     "typescript": "^5.7.2",
     "zod": "^3.24.1"
   },
   "dependencies": {
-    "@api3/chains": "^11.8.0",
     "ethers": "^6.13.2",
+    "viem": "^2.23.2",
```
bbenligiray, line 105: Two big users of this package, market and auctioneer depend on this, so it's not a big deal imo

bbenligiray:
Since this repo now uses its own chain-stuff instead of importing it from `@api3/chains`, `pnpm build` needs to be run before virtually everything (which is fine?)

src/hardhat-config.test.ts
bbenligiray: Yeah I suppose so

hardhat.build.config.ts
```diff
@@ -0,0 +1,93 @@
+import * as fs from 'node:fs';
```
hiletmis, line 1: This file is being used to configure hardhat now, but why not just use `hardhat.config.ts` instead

package.json
```diff
@@ -32,14 +32,17 @@
     "api3-contracts": "./dist/bin/cli.js"
   },
   "scripts": {
-    "build": "pnpm check && hardhat compile && pnpm generate && tsc -p tsconfig.build.json",
-    "check": "pnpm check:deployment-addresses && pnpm check:deployment-config",
+    "build": "pnpm build:hardhat && pnpm check  && pnpm generate && tsc -p tsconfig.build.json",
```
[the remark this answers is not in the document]
bbenligiray: In api3dao/chains `validate` (our `check` equivalent) is not a part of `build`, instead it's done in prepush hook and CI. However, I don't see a downside to run `check` in `build` here, and the change to the order makes sense.

hardhat.build.config.ts
bbenligiray: Because `hardhat.config.ts` imports `hardhatConfig` from `./src/index`, which needs `./typechain-types` to have been generated, which needs to run `hardhat compile`, which depends on `hardhat.config.ts`. So I removed the sections of `hardhat.config.ts` that needed `hardhatConfig`, named it `hardhat.build.config.ts`, and ran `hardhat compile` with that. Also related https://github.com/api3dao/contracts/issues/396

hiletmis (approved):

PR #402, opened 2025-02-20
Change dApps format

Siegrift:
This PR changes the dApps format to support multiple aliases per dApp. Each alias can be set for multiple chains. While doing this breaking change, I've also addressed the following:
1. Only listed dApps relevant for OEV integrations and removed the rest
2. Updated the dApps according to our internal dApp management sheet. If a dApp is on multiple chains, I only used the chains that are relevant for the short term
3. Updated dApp names to match Defillama and enforced the filename to match the dApp name. This is also nice, in case the dApp rebrands in the future - we can rename the name and filename, but leave the aliases as is
4. Enhanced proxy CLI checks and generation script checks


bbenligiray:
> Updated dApp names to match Defillama and enforced the filename to match the dApp name

The dApp names were set to satisfy dApp preference based on their branding (+guesswork when there were inconsistencies) because at the very least these names will appear on the proxy page of the market (if not other user-facing frontends) and the dApps would appreciate us calling them what they call themselves. DefiLlama is not much of a factor.

bbenligiray:
> This is also nice, in case the dApp rebrands in the future - we can rename the name and filename, but leave the aliases as is

I don't see how having to rename the file when the dApp rebrands is nice, "we don't need to rename the file even if the dApp rebrands" sounds nicer.

Related to this, it kinda sucks that there is no immutable way to address a dApp anymore. How do I find the specific dApp that I'm looking for under `DAPPS`, go through `aliases` of every item? Am I guaranteed that aliases are not duplicated across dApps?

data/dapps/dtrinity.json
```diff
@@ -1,5 +1,7 @@
 {
-  "alias": "dtrinity",
-  "name": "dTrinity",
+  "aliases": {
+    "dtrinity": ["fraxtal"]
+  },
+  "name": "dTRINITY",
```
bbenligiray, line 5: This one seems to have changed since I looked at it

bbenligiray:
> Only listed dApps relevant for OEV integrations and removed the rest

A lot of the dApps that I've added have practically filled the form. You didn't remove the ones that are in the BD pipeline right?

data/dapps/orbit-protocol.json
```diff
@@ -0,0 +1,7 @@
+{
+  "aliases": {
+    "orbit": ["blast"]
+  },
+  "name": "Orbit Protocol",
```
bbenligiray, line 5: These guys for example call themselves Orbit everywhere except Twitter so I went with the version without fluff

scripts/generate-dapps.ts
```diff
@@ -18,6 +20,10 @@ const HEADER_CONTENT = `// =====================================================
 import { type Dapp } from '../types';
 `;
 
+function toImprovisedKebabCase(value: string) {
```
bbenligiray, line 23: There is this in the same repo https://github.com/api3dao/contracts/blob/main/src/cli.ts#L40

Siegrift: I'll copy paste it, because I don't want to create an utils module just for this function.

data/dapps/dtrinity.json
Siegrift: Yeah, as I mentioned - I've renamed them according to Defillama. But they're using this capitalization also on their homepage.

Siegrift (commented):
> The dApp names were set to satisfy dApp preference based on their branding (+guesswork when there were inconsistencies) because at the very least these names will appear on the proxy page of the market (if not other user-facing frontends) and the dApps would appreciate us calling them what they call themselves. DefiLlama is not much of a factor.

My assumption is that what they have in Defillama is acceptable to them. It may not always be the ideal though (I think a good example is "Orbit Protocol" which is probably not called "Orbit" because there is a lot of stuff named Orbit in Defillama)... A big advantage of using Defillama is that it's a convention we can follow instead of bikeshed the dApp name. We can also implement automated check to verify we don't miss a rebrand. There is still going to be guesswork for dApps not included in Defillama, but another assumption is that all "worthy" dApps are going to be listed there at some point.

> I don't see how having to rename the file when the dApp rebrands is nice, "we don't need to rename the file even if the dApp rebrands" sounds nicer.

I disagree. It would be pretty weird if we had filename "hana-finance" and dApp name would be "Lalala Lend". I think when rebrand occurs, we may even want to preserve the old name (in some separate field). I think some UI consumers may want to show that information (e.g. show current title and also "previously known as XYZ").

> How do I find the specific dApp that I'm looking for under DAPPS, go through aliases of every item?

I guess you mean "how to determine a dApp given an alias". Yes, in that case you traverse this list. You are guaranteed these will be unique (this is obviously something we want to enforce and the checks are already implemented).

> Even if this was properly enforced, having the unique identifiers in separate groups feels unintuitive.
For example, say I'm searching for init, new @api3/contracts got published, and I want my bot to notify me if Init has expanded to a new chain. I'll have to do DAPPS.find(dapp => dapp.aliases.includes('init')), which feels awkward to be tbqhf. I'm not sure if there is a solution though.

The alternative would be to change the way we store dApps data. Basically we could have a one big JSON object with keys `dApps` and `aliases`. The `aliases` would be a record that define a chain and pointer to a dApp. I wouldn't go that way, but it's an option.

> A lot of the dApps that I've added have practically filled the form. You didn't remove the ones that are still in the BD pipeline right?

Yes, given the context I have. I listed all of the dApps that were in the dApp management sheet. I guess best for @Ashar2shahid to confirm whether there is some dApp that is missing in the sheet and we should have an alias for.

bbenligiray:
> My assumption is that what they have in Defillama is acceptable to them. It may not always be the ideal though (I think a good example is "Orbit Protocol" which is probably not called "Orbit" because there is a lot of stuff named Orbit in Defillama)... A big advantage of using Defillama is that it's a convention we can follow instead of bikeshed the dApp name. We can also implement automated check to verify we don't miss a rebrand. There is still going to be guesswork for dApps not included in Defillama, but another assumption is that all "worthy" dApps are going to be listed there at some point.

I think it's common for dApps to completely forget about DefiLlama after the initial integration (case in point, I don't see DefiLlama [here](https://docs.google.com/document/d/1lLJO0v_PE2jQ_5e-j4fd9zgS0A3oTV_dadnGJSNDjo4/edit?tab=t.0#heading=h.rwc6qlzacrxp)). Yours is still a passable approach though, and obviously more objective and automateable.

> I disagree. It would be pretty weird if we had filename "hana-finance" and dApp name would be "Lalala Lend".

We never had any problems with [`bsc`](https://github.com/api3dao/contracts/blob/main/data/chains/bsc.json) but granted, that's only used internally (but arguably, the user will also never see the dApp file names). I don't feel great about using two different conventions in the same repo though.

> I think when rebrand occurs, we may even want to preserve the old name (in some separate field). I think some UI consumers may want to show that information (e.g. show current title and also "previously known as XYZ").

The dApp itself certainly wouldn't. Rebranding is difficult enough as is, you don't want to deal with being dead-named.

> the checks are already implemented

Ah I missed that. 

> The alternative would be to change the way we store dApps data. Basically we could have a one big JSON object with keys dApps and aliases. The aliases would be a record that define a chain and pointer to a dApp. I wouldn't go that way, but it's an option.

Yeah that occurred to me but I was too lazy to type it out. I agree that the current thing is preferable to that.

bbenligiray (approved):
Looks good, except it may be removing dApps that it shouldn't. This won't disrupt anything on-chain, and adding them back wouldn't be too difficult.

bbenligiray:
btw you should add a major changeset

Siegrift:
> Looks good

Actually, not quite - both market and payout reports need to know be able to differentiate aliases on the same chain. Currently there is a lack of such display name. A nice way to solve this is to add a `tag` property for an alias. For Compound V3, the file would look like:
```
{
  "aliases": {
    "compound-v3-ether": { "chains": ["ethereum", "optimism", "base", "arbitrum"], "tag": "ETH market" },
    "compound-v3-usde": { "chains": ["mantle"], "tag": "USDe market" }
  },
  "name": "Compound V3",
  "homepageUrl": "https://v3-app.compound.finance/"
}
```
The `tag` is name generic on-purpose, because protocols name the markets differently - e.g. Morhpo has "vaults", Silo "silos", etc...

> except it may be removing dApps that it shouldn't. This won't disrupt anything on-chain, and adding them back wouldn't be too difficult.

Yeah, easy enough to add so I'd rather use this opportunity and clear this up now.

> Yours is still a passable approach though, and obviously more objective and automateable.

I think it's a good enough baseline to follow, especially for new integrations. We can make exception if asked by the dApp.

> We never had any problems with [bsc](https://github.com/api3dao/contracts/blob/main/data/chains/bsc.json)

I think that's because `bsc` is still good name. Better example is `x-dai` which got rebranded to `gnosis`.

> btw you should add a major changeset

Done. I'll wait for your review though, since I've added non-trivial changes.

data/dapps/silo-finance.json
```diff
@@ -0,0 +1,6 @@
+{
+  "aliases": {
+    "silo-finance-v1-rdnt": { "chains": ["arbitrum"], "title": "Silo Finance RDNT market" }
```
Siegrift, line 3: According to [Defillama](https://defillama.com/protocol/silo-finance), the dApp name is `Silo Finance` but it has two versions (for which Defillama has subpages).

For maximum clarity I derived the dApp alias from the dApp name, the version of the protocol and the market asset.

Siegrift (commented):
Hopefully the final changes. What I've done:
1. Added a property `title` to the alias record and removed the `name` from the dApp part.
2. Not added the `description` property, because it can't provide enough context. E.g. given a multi-vault dApp, the API3 market needs to convey to the user that it only shows proxies for a single vault (and user may want need to register other vaults first).
3. Added `multiAliased` option to the dApp record to signal these special apps. It's up to the consumers to decide what to do with this information.

I remembered why I wanted to preserve the `name` in the file. It provides the canonical name for the dApp, that provides more context about the dApp. The homepage URL is optional, so I figured we want to keep at least the name. I've also used it to derive check the filename naming. That said, I think moving the name (title) to the alias section is better. I was also thinking of making the homepage URL required, but decided to leave it optional for more flexibility.

Siegrift (commented):
Added https://github.com/api3dao/contracts/pull/402/commits/5de786f8958e0c3bb2387b8c0613c7f277b0b590 with support for description field and CLI. 

It was a bit difficult to create a description that would both fit the market and the CLI, but I think the current works well enough for both.

Running `pnpm run cli print-api3readerproxyv1-address --dapp-alias silo-finance-v1-rdnt --chain-id 42161 --dapi-name "RDNT/USD" --strict` prints:

image

data/dapps/silo-finance.json
```diff
@@ -0,0 +1,11 @@
+{
+  "aliases": {
+    "silo-finance-v1-rdnt": {
+      "chains": ["arbitrum"],
+      "title": "Silo Finance RDNT market",
+      "description": "The dAPI proxies specific for Silo Finance RDNT market on Arbitrum. Proxies for other Silo Finance markets should use different configuration."
```
Siegrift, line 6: Alternatively:
```suggestion
      "description": "Silo Finance consists of multiple independent markets. Each market should use separate OEV proxies. This OEV proxy is intended to be used only for Silo Finance RDNT market on Arbitrum."
```

Btw. using the term `market` because they also use that on their page -> https://app.silo.finance/ (although they are referring to these markets as silos elsewhere).

src/proxy.ts
```diff
@@ -54,9 +58,18 @@ function computeCommunalApi3ReaderProxyV1Address(chainId: ethers.BigNumberish, d
 }
 
 function computeDappId(dappAlias: string, chainId: ethers.BigNumberish) {
-  if (!DAPPS.some((dapp) => dapp.alias === dappAlias)) {
+  const dApp = DAPPS.find((dapp) => Object.hasOwn(dapp.aliases, dappAlias));
+  if (!dApp) {
     throw new Error(`dApp with alias ${dappAlias} not registered to the package`);
   }
+  const chainAlias = CHAINS.find((chain) => chain.id === chainId)?.alias;
```
bbenligiray, line 65: You shouldn't be strict about this stuff in here, you're breaking the non-strict mode of the CLI command

data/dapps/compound-finance.json
```diff
@@ -1,5 +1,11 @@
 {
-  "alias": "compound-finance",
-  "name": "Compound",
+  "aliases": {
+    "compound-finance-usde": {
+      "chains": ["mantle"],
+      "title": "Compound Finance USDe market",
+      "description": "The dAPI proxies specific for Compound Finance USDe market on Mantle. Proxies for other Compound Finance markets should use different configuration."
```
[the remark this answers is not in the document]
bbenligiray: I don't see the point in referring to the dApp name in these descriptions.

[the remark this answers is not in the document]
bbenligiray: I would avoid being wordy, like `Only to be used for USDe markets.`

data/dapps/silo-finance.json
bbenligiray: `Only to be used for the RDNT market.` I don't see it omitting any information from the above. A URL would further clarify things but that's because multiple aliases need their own `homepageUrl` fields.

data/dapps/compound-finance.json
```diff
@@ -1,5 +1,11 @@
 {
-  "alias": "compound-finance",
-  "name": "Compound",
```
bbenligiray, line 3: You still need a dApp name if you will have the user first select the dApp and then the subselection for example on the Market frontend.

bbenligiray (approved):
I find `aliases` and `multiAliased` ridiculous

data/dapps/silo-finance.json
Siegrift: > `Only to be used for the RDNT market. `

Changed.

> A URL would further clarify things but that's because multiple aliases need their own homepageUrl fields.

Yeah, that would further help to distinguish those, and we could have something like this:
```jsonc
{
 "aliases": {
   "silo-finance-v1-rdnt": {
     "title": "Silo Finance RDNT market",
     "description": "Only to be used for the RDNT market.",
     "chains": {
       "arbitrum": {
         "homepageUrl": "https://app.silo.finance/silo/0x19d3F8D09773065867e9fD11716229e73481c55A"
       }
     }
   }
 }
}
```

the problem with that is that such URL is not always available (for example [Moonwell](https://moonwell.fi/markets) doesn't have an URL for a specific chain deployment). We could special case this, but I think showing only the description is good enough.

data/dapps/compound-finance.json
```diff
@@ -1,5 +1,11 @@
 {
```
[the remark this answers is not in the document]
Siegrift: I'm trying to follow [Defillamma](https://defillama.com/protocol/compound-finance) whenever possible to avoid unnecessary discussions on what should be the alias and the filename. The subjective choice is how to differentiate the markets and for that I've used a suffix `-usde`.

data/dapps/aurum.json
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
Siegrift: Yeah, [me too](https://defillama.com/protocol/aurum), but we've already deployed the proxies for them and the integration is in-progress so it's better to leave this as is.

data/dapps/machfi.json
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
Siegrift: Already integrated.

data/dapps/compound-finance.json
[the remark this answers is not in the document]
Siegrift: Changed.

src/proxy.ts
Siegrift: I added it for consistency 🤔 Shouldn't we then also remove the following?
```
  const dApp = DAPPS.find((dapp) => Object.hasOwn(dapp.aliases, dappAlias));
  if (!dApp) {
    throw new Error(`dApp with alias ${dappAlias} not registered to the package`);
  }
```

For now, removed all of those checks, but can revert in case I'm missing something.

data/dapps/compound-finance.json
Siegrift: My thinking was that the market UI will remain the same and the OEV Rewards select field will contain all of the relevant `title` options. I expect it will take some time until this becomes too messy and I'd defer subselection to a later stage.

data/dapps/compound-finance.json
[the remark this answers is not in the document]
bbenligiray: Right okay, I remember that discussion. I don't have an issue with `-usde`.

src/proxy.ts
bbenligiray: I agree on removing that too. Since we're doing a major change, the function can be renamed as `unsafeComputeDappId()` to mitigate potential issues because of the lack of checks.

data/dapps/enclabs.json
```diff
@@ -1,5 +1,6 @@
 {
-  "alias": "enclabs",
-  "name": "Enclabs",
-  "homepageUrl": "https://x.com/EnclabsFi"
+  "aliases": {
+    "enclabs": { "chains": ["sonic"], "title": "Enclabs" }
+  },
+  "homepageUrl": "https://www.enclabs.finance/#/?chainId=146"
```
bbenligiray, line 5:
```suggestion
  "homepageUrl": "https://www.enclabs.finance/
```

data/dapps/takara-lend.json
```diff
@@ -0,0 +1,6 @@
+{
+  "aliases": {
+    "takara": { "chains": ["sei"], "title": "Takara Lend" }
+  },
+  "homepageUrl": "https://x.com/takaralend"
```
bbenligiray, line 5:
```suggestion
  "homepageUrl": "https://takaralend.com/"
```

data/dapps/vicuna-finance.json
```diff
@@ -1,5 +1,6 @@
 {
-  "alias": "vicuna-finance",
-  "name": "Vicuna Finance",
+  "aliases": {
+    "vicuna-finance": { "chains": ["sonic"], "title": "Vicuna Finance" }
+  },
   "homepageUrl": "https://x.com/VicunaFinance/"
```
bbenligiray, line 5:
```suggestion
  "homepageUrl": "https://vicunafinance.com/"
```

data/dapps/silo-finance.json
bbenligiray: Yeah, I wouldn't change the schema further. The `dapis.json` approach is to use such links in the description where applicable.

src/proxy.ts
bbenligiray: Huh https://github.com/api3dao/data-feed-reader-example/blob/31902685bc20e192069d5072393e925796366568/scripts/deploy-dapp-specific-api3readerproxyv1-unsafe.js#L11

Siegrift: I was afraid for a moment that I messed up something with the rename. Do you want me to fire up a PR in data-feed-reader-example which uses this function from contracts?

bbenligiray: Nah, it's just that it looks like I was deliberate with the dApp check

PR #429, opened 2025-03-16
Add check for chain support ordering

### LINKED ISSUE #389: Verify alphabetical ordering in `data/chain-support.json`
It's currently not satisfied

dcroote:
Closes #389. I checked it works by confirming the script correctly fails before the second commit (since the arrays are currently not sorted) and passes after the second commit. Script (logging) approach similar to [scripts/check-chains.ts](https://github.com/api3dao/contracts/blob/main/scripts/check-chains.ts).


bbenligiray (approved):
Looks good, thanks

PR #430, opened 2025-03-17
Update RPC provider configurations

bdrhn9:
I've divided the changes into multiple commits to simplify the review process. The primary goal of this PR is to increase the number of RPC providers to 4-5 while prioritizing reliability. Unreliable providers have been either removed or replaced with better alternatives.


.changeset/puny-toes-leave.md
```diff
@@ -0,0 +1,19 @@
+---
+'@api3/contracts': patch
```
bbenligiray, line 2: :+1:

bbenligiray (approved):

PR #452, opened 2025-04-14
Csur-2025-04-14

hiletmis:
Related to https://github.com/api3dao/tasks/issues/1680 and https://github.com/api3dao/data-feeds/issues/2747

`hardhat-deploy` is downgraded to `0.14.0` as `0.14.1` is an accidental `major` release and has some bugs that prevent contract deployment. The problem seems to be [fixed](https://github.com/wighawag/hardhat-deploy/commit/9de768960d0253ca84a7dcb568a77080867d4196) but I decided to downgrade it to a known version.

Contract verification is a bit odd on this chain. There is no API. They require users to add contract ABIs to [this repo](https://github.com/princesinha19/megaeth-abis)


data/chains/mega-testnet.json
```diff
@@ -0,0 +1,17 @@
+{
+  "alias": "mega-testnet",
+  "decimals": 18,
+  "explorer": {
+    "browserUrl": "https://www.megaexplorer.xyz/"
+  },
+  "id": "6342",
+  "name": "MEGA testnet",
```
bbenligiray, line 8: I see MegaETH (testnet) a lot more than MEGA testnet

bdrhn9 (commented):
I'm generally okay with this chain. However, I delved into the details of the **[mini blocks](https://docs.megaeth.com/mini-blocks)** feature to investigate whether it could cause issues for Airseeker. From my notes and understanding:  

- Only RPC methods that use `blockTag` can access mini blocks. So, when you set `blockTag` to `latest` or `pending`, you're effectively reading from mini blocks.  
- In Airseeker, we fetch on-chain data and block numbers simultaneously using `eth_call(blockTag: latest)` and `eth_blockNumber`, respectively.  
  - Since `eth_call(blockTag: latest)` pulls data from a mini block while `eth_blockNumber` returns the original EVM block number, a slight discrepancy exists.  
  - However, this discrepancy doesn’t trigger redundant updates, making it acceptable.  
  - That said, I can’t rule out potential finality issues at this stage.  
  - Basically, these mini blocks sound like pending transactions in the tx pool that are almost certain to be confirmed by the sequencer.

bbenligiray (approved):

PR #456, opened 2025-04-16
add segment,nerite to dapps

Ashar2shahid:


Siegrift:
Thanks for noticing the generated dApps not being included.

bbenligiray:
Welp, it was also missing the changeset

PR #468, opened 2025-04-23
Add Malda on Linea to dApp details

Siegrift:


data/dapps/malda.json
```diff
@@ -0,0 +1,6 @@
+{
+  "aliases": {
+    "malda": { "chains": ["linea"], "title": "Malda" }
```
Siegrift, line 3: There is no Defillama page, but they use `Malda` consistently in their webiste, logo and twitter.

Issue #388, opened 2025-02-14
`ronin` contracts are deployed non-deterministically


bbenligiray:
We prefer deploying the contracts deterministically for [local bytecode verification](https://github.com/api3dao/contracts/blob/main/scripts/verify-deployments.ts) to have lower RPC requirements (see https://github.com/api3dao/contracts/issues/223). However, `ronin` blacklisted the CREATE2 deployer contract at launch (plalnned to be undone months later) and we had to integrate urgently to meet specific user demand. As a result, contracts on `ronin` specifically are deployed non-deterministically.

PR #472, opened 2025-04-25
Adds proxies composite and wrapper contracts

### LINKED ISSUE #378: Implement proxy combinators
We should have a basic proxy combinators to support combined feeds for dApps conveniece. Such proxies need to be verifiable by the CLI.

acenolaza:
Closes #378 

### Changes summary

1. Adds a new composite proxy contract to return the multiplication of values read from 2 other Api3 proxies.
2. Adds a new proxy to return the inverse rate using  the reciprocal function.
3. Adds a new proxy that returns the normalized to 18 decimals value read from a Chainlink aggregator
4. Adds an Chainlink aggregator adapter contract that can be used to read any Api3 proxy contract and scale its value to decimals from 1 to 36.
5. Adds a Chainlink aggregator base contract for Api3 proxies to inherit.

Similarly to [API3CombineProxies](https://explorer.mantle.xyz/address/0xd5C92f854E3d0e14DEB5c367253D3e62383c2d50), multiple Api3 proxy contracts can be compounded for more complex rates.


bbenligiray:
It doesn't seem to support OEV. I would prefer a non-upgradeable proxy to two Api3ReaderProxyV1 contracts.

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
```diff
@@ -0,0 +1,195 @@
...
+            crossRate = (val2 * 1e18) / val1;
+        } else if (calculationType == CalculationType.Multiply) {
+            crossRate = (val1 * val2) / 1e18;
+        }
+        if (crossRate > type(int224).max || crossRate < type(int224).min) {
+            revert ValueOutOfRange(crossRate);
+        }
+
+        value = int224(crossRate);
+        timestamp = timestamp1 < timestamp2 ? timestamp1 : timestamp2;
```
bbenligiray, line 101: `block.timestamp` imo

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /// @dev Functions that use the round ID as an argument are not supported
+    function getTimestamp(uint256) external pure override returns (uint256) {
+        revert FunctionIsNotSupported();
+    }
+
+    /// @dev API3 feeds always use 18 decimals
+    function decimals() external pure override returns (uint8) {
+        return 18;
```
bbenligiray, line 143: I wonder if there are cases where the user requires different decimals

contracts/api3-server-v1/proxies/interfaces/IApi3CrossRateReaderProxyV1.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+import "../../../interfaces/IApi3ReaderProxy.sol";
+
+interface IApi3CrossRateReaderProxyV1 is IApi3ReaderProxy {
+    enum CalculationType {
+        Divide1By2,
+        Divide2By1,
```
bbenligiray, line 9: I don't get why both are needed

contracts/api3-server-v1/proxies/interfaces/IApi3CrossRateReaderProxyV1.sol
```diff
@@ -0,0 +1,28 @@
...
+        Multiply
```
bbenligiray, line 10: Is there an existing need?

contracts/api3-server-v1/proxies/interfaces/IApi3CrossRateReaderProxyV1.sol
acenolaza: Mainly to support more cross rate combinations. Say you have proxy1 = ETH/USD and proxy2 = SOL/USD. If we wanted a cross rate for ETH/SOL we would divide ETH/USD by SOL/USD but if we wanted SOL/ETH then it would be SOL/USD divided by ETH/USD. An alternative could be to use the reciprocal of the proxies but I wanted to keep it as simple as possible and avoid having to deal with too many if statements based on base and quote checks 🤷🏻

contracts/api3-server-v1/proxies/interfaces/IApi3CrossRateReaderProxyV1.sol
acenolaza: I was able to only find cases where we multiply the proxies tbh 🤔

acenolaza:
> It doesn't seem to support OEV.

> I would prefer a non-upgradeable proxy to two Api3ReaderProxyV1 contracts.

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: I think that as long as proxies are API3 proxies then then answer is no

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
```diff
@@ -0,0 +1,195 @@
...
+        int256 val2 = int256(value2);
+        int256 crossRate;
+        if (calculationType == CalculationType.Divide1By2) {
+            if (val2 == 0) revert ProxyReturnedZero(proxy2);
+            crossRate = (val1 * 1e18) / val2;
+        } else if (calculationType == CalculationType.Divide2By1) {
+            if (val1 == 0) revert ProxyReturnedZero(proxy1);
+            crossRate = (val2 * 1e18) / val1;
+        } else if (calculationType == CalculationType.Multiply) {
+            crossRate = (val1 * val2) / 1e18;
```
acenolaza, line 94: One thing I've just realized is that this can overflow if `val1`, `val2` or both are too large (unlikely if these are going to always be asset prices). Math library from Openzeppelin provides a `mulDiv() `function that would fix this issue but here we are dealing with `int` and the function only supports `uint` so we could maybe wrap the `mulDiv()` with a helper function that converts `int` to `uint` like this:

```
    function mulDivSigned(
        int256 x,
        int256 y,
        int256 denominator
    ) internal pure returns (int256 result) {
        bool negativeResult = (x ^ y ^ denominator) < 0;

        uint256 ax = x < 0 ? uint256(-x) : uint256(x);
        uint256 ay = y < 0 ? uint256(-y) : uint256(y);
        uint256 ad = denominator < 0
            ? uint256(-denominator)
            : uint256(denominator);

        uint256 r = Math.mulDiv(ax, ay, ad);
        require(r <= uint256(type(int256).max), "mulDivSigned overflow");

        result = negativeResult ? -int256(r) : int256(r);
    }
```

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
```diff
@@ -0,0 +1,195 @@
...
+        }
+        if (crossRate > type(int224).max || crossRate < type(int224).min) {
+            revert ValueOutOfRange(crossRate);
+        }
```
acenolaza, line 98: I'm also now thinking that this check is redundant because there is no way a multiplication that does not overflow `int256` could result in a value that does overflow `int224` after being divided by `1e18`

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
```diff
@@ -0,0 +1,193 @@
...
+/// that it is deployed by Api3CrossRateReaderProxyV1Factory, the owner will be
+/// the owner of Api3CrossRateReaderProxyV1Factory at the time of deployment.
+/// @dev For a gas-cheap `read()` implementation, this upgradeable contract
+/// uses immutable variables (rather than initializable ones). To enable this,
+/// an Api3CrossRateReaderProxyV1 needs to be deployed for each unique
+/// combination of variables. The end user does not need to concern themselves
+/// with this, as Api3CrossRateReaderProxyV1Factory abstracts this detail away.
+/// Refer to https://github.com/api3dao/migrate-from-chainlink-to-api3 for more
+/// information about the Chainlink interface implementation.
+contract Api3CrossRateReaderProxyV1 is
```
Siegrift, line 23: I'm not a fan of "cross rate" in the name. I think `CompositeApi3ReaderProxyV1` would be better.

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
```diff
@@ -0,0 +1,193 @@
...
+    UUPSUpgradeable,
+    OwnableUpgradeable,
+    AggregatorV2V3Interface,
+    IApi3CrossRateReaderProxyV1
+{
+    /// @notice First IApi3ReaderProxy contract address
+    address public immutable override proxy1;
```
Siegrift, line 30: Note that there is also a case when there is a non Api3ReaderProxy feed that we need to compose with. I think we should require `AggregatorV2V3Interface` support, but it would be nice to have a contract that allows wrapping a feed reader supporting `AggregatorV2V3Interface` that scales it to 1e18 so that we can support such case using  this wrapper.

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
```diff
@@ -0,0 +1,193 @@
...
+    address public immutable override proxy2;
+
+    /// @notice Calculation type to be used for the cross rate
+    CalculationType public immutable override calculationType;
+
+    /// @notice Derived dAPI name of the cross rate as a bytes32 string
+    /// @dev For example if the two proxies are "USD/USDT" and "USDT/EUR", the
+    /// derived dAPI name should be "USD/EUR". This is just for informational
+    /// purposes and is not used in any calculations
+    bytes32 public immutable override crossRateDapiName;
```
Siegrift, line 42: I don't really like it. This creates a friction point during deployment of the reader proxy and it's something that can go wrong.

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
```diff
@@ -0,0 +1,193 @@
...
+    constructor(
+        address proxy1_,
+        address proxy2_,
+        CalculationType calculationType_,
+        bytes32 crossRateDapiName_
+    ) {
+        proxy1 = proxy1_;
+        proxy2 = proxy2_;
+        calculationType = calculationType_;
+        crossRateDapiName = crossRateDapiName_;
```
Siegrift, line 58: There are only two combinator types - multiple and divide. I'd prefer two separate contracts for that, maybe `ProductCompositeApi3ReaderProxyV1` and `QuotientCompositeApi3ReaderProxyV1`

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
Siegrift: There is no good aggregation method, so we can go with `block.timestamp` as well. What I have against it is that it's quite different from the timestamp of `Api3ReaderProxyV1` which feels weird. Also, this also doesn't address the [timestamp check problem](https://docs.api3.org/dapps/integration/contract-integration.html#using-timestamp) that complicates OEV. That said, if we go ahead with `block.timestamp` now, we can more easily migrate the standard proxies in the future.

contracts/api3-server-v1/proxies/interfaces/IApi3CrossRateReaderProxyV1.sol
Siegrift: I also don't get this. You can change the divisor based on the order of the proxies.

contracts/api3-server-v1/proxies/interfaces/IApi3CrossRateReaderProxyV1.sol
Siegrift: I think there was also a case when a dApp used division, but I can't remember which one it was. It probably wasn't for our feeds though.

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/cd638f30c3fd424d1162d84d839c79ef02f405cc

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/7c58ea532f319360ee8bee22982ec3b072a32f49

contracts/api3-server-v1/proxies/interfaces/IApi3CrossRateReaderProxyV1.sol
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/8e22dff085862c172358f0b5a31def4e6e6ec06d

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/29cd894ae47e55c70a6fa057c07f5c52e536bc2f

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
bbenligiray: I don't think any user contract was ever built anticipating to use Api3 proxies so they may have had to build an adapter for decimal conversion

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
bbenligiray: Fwiw there is also the `1/proxy` case

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
bbenligiray: Then I would just use `latestAnswer()` and not expect the `read()` interface

bbenligiray:
I think the factory contract is redundant here. You can just deploy CompositeApi3ReaderProxyV1 deterministically to the same effect.

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: It seems that I should only write a single `ProductCompositeApi3ReaderProxyV1` for now 🤔

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: I have only found cases like [this](https://seitrace.com/address/0x3e10e6eF3c602885a9Cafd9785BD2143c52fbFF1?chain=pacific-1&contract=read_contract&slug=&tab=contract) where one of the proxies is not an `IApi3ReaderProxy` (or legacy `IProxy`) and they add a wrapped that implements a `read()` method with the same [signature](https://seitrace.com/address/0xaFb041c4268Efbf217ACEea6D795FA16A3b31A2D?chain=pacific-1&tab=contract) as the one in `IApi3ReaderProxy` interface.

contracts/api3-server-v1/proxies/interfaces/IApi3CrossRateReaderProxyV1.sol
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/5512753d18a40b90ee9ac520bfeb5b3efe295862

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: I think I might have found a case where they use a wrapper to convert the value to 8 decimals https://fraxscan.com/address/0x68E1085A7639e80257f7f159A611Dd944278A24e#code#F12#L46

acenolaza:
> > I think the factory contract is redundant here. You can just deploy CompositeApi3ReaderProxyV1 deterministically to the same effect.
> 
> Wouldn't that force us to always use the same deployer EOA?

https://github.com/api3dao/contracts/pull/472/commits/da2919d56eec34e651568bcf8dbed42353f624b0

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/012d779b03e408bde97ef75e23a558c698deab1d

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/012d779b03e408bde97ef75e23a558c698deab1d

acenolaza: Related to https://github.com/api3dao/contracts/pull/472#discussion_r2070480945

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
Siegrift: > Fwiw there is also the 1/proxy case

That can be implemented using a `ProductCompositeApi3ReaderProxyV1` and a contract that returns scalar value 1. This is also something we see being used and might add now.

contracts/api3-server-v1/proxies/CompositeApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,154 @@
...
+    /// unsuitable for an int256. The timestamp is the current block timestamp
+    /// @return value Value of the rate composition
+    /// @return timestamp Timestamp of the rate composition
+    function read()
+        public
+        view
+        override
+        returns (int224 value, uint32 timestamp)
+    {
+        (int224 value1, ) = IApi3ReaderProxy(proxy1).read();
```
Siegrift, line 63: If we go with IApi3ReaderProxy requirement here I think it makes sense to hardcode 18 for decimals in the implementation. I like that the implementation is very clean this way. If we want to be generic, we can expect `latestAnswer` as Burak pointed out, but I prefer the current implementation.

contracts/api3-server-v1/proxies/CompositeApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,154 @@
...
+    /// @notice First IApi3ReaderProxy contract address
+    address public immutable override proxy1;
+
+    /// @notice Second IApi3ReaderProxy contract address
+    address public immutable override proxy2;
+
+    /// @notice Decimals of the rate composition
+    uint8 public immutable override decimals;
+
+    uint256 private immutable baseUnit;
```
Siegrift, line 26: This doesn't work. In the implementation you note `IApi3ReaderProxy always return values in 18 decimals` but here you allow the proxies to have different decimals.

I think there should be a separate contract for scaling the decimals precision. That way you can combine the proxies and at the end wrap it in `ScaledApi3FeedProxyV1` (I'd intentionally avoid choose a different naming to make it more apparent this is not compatible with proxies). I guess this one could depend on `AggregatorV2V3Interface` and be used to convert the non-Api3 proxies to `AggregatorV2V3Interface`.

contracts/api3-server-v1/proxies/CompositeApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,154 @@
...
+import "hardhat/console.sol";
+
+/// @title An immutable proxy contract that is used to read a composite data feed
+/// by multiplying two IApi3ReaderProxy data feeds
+/// @dev This contract implements the AggregatorV2V3Interface to be compatible with
+/// Chainlink aggregators. This allows the contract to be used as a drop-in
+/// replacement for Chainlink aggregators in existing dApps.
+/// Refer to https://github.com/api3dao/migrate-from-chainlink-to-api3 for more
+/// information about the Chainlink interface implementation.
+contract CompositeApi3ReaderProxyV1 is ICompositeApi3ReaderProxyV1 {
```
Siegrift, line 16: What about the other use cases? My thinking is to have a complete proxy composition library. Even if there is no use case now, I want to avoid having another iteration like this when the use case happens.

contracts/api3-server-v1/proxies/CompositeApi3ReaderProxyV1.sol
acenolaza: This is a valid point, I should have said that `Api3ReaderProxyV1` proxies always return 18 decimals but it still breaks when either `proxy1` or `proxy2` don't return 18 decimals. I guess we can't even expect `proxy1` and `proxy2` to be `AggregatorV2V3Interface` because they might even be old legacy `IProxy` proxies 🤔 I guess having a wrapper scaling contract to be used at the end sounds like a good idea 👍🏻

acenolaza: Another alternative could be to set the decimals for each proxy via constructor arguments and keep the `baseUnit` scaling

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
Siegrift: I was thinking more about
> That said, if we go ahead with block.timestamp now, we can more easily migrate the standard proxies in the future.

and it feels to me we could do it now. If it's OK to have `block.timestamp` in this contract, it is OK to have it also for the proxies and we can have create `IApi3ReaderProxyV2` that also returns the `block.timestamp`. The common user complaint would be able about the staleness check, but the workaround here is to still expose the off-chain timestamp the users could validate.

WDYT? (This is still a task independent to this and should be addressed separately).

acenolaza: Sounds good 👍🏻 I can tackle this in a separate issue and implement an `Api3ReaderProxyV2` that returns `block.timestamp` and maintains backward compatibility with `IApi3ReaderProxy` to not break current dApp integrations.

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: We can now wrap `AggregatorV2V3Interface` with [NormalizedApi3ReaderProxyV1](https://github.com/api3dao/contracts/pull/472/commits/2fe44e816ce20ff0775324bd86b71f2b510d25c7)

contracts/api3-server-v1/proxies/CompositeApi3ReaderProxyV1.sol
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/0cb5750059430cf8fc86786f89a58dff8920b6a8

contracts/api3-server-v1/proxies/Api3CrossRateReaderProxyV1.sol
acenolaza: * Renamed to [ProductApi3ReaderProxyV1](https://github.com/api3dao/contracts/pull/472/commits/fd6b0832f0237a8eab7023dea781e7194dfeac42) to keep it short
* Added [InverseApi3ReaderProxyV1](https://github.com/api3dao/contracts/pull/472/commits/cedc8ab95d84589156f2f7a87d99508fc62ad421)

contracts/api3-server-v1/proxies/CompositeApi3ReaderProxyV1.sol
bbenligiray: > My thinking is to have a complete proxy composition library

FWIW I regretted every time I broke a contract down into multiple contracts

bbenligiray:
> Wouldn't that force us to always use the same deployer EOA?

No, the deterministic deployment doesn't care about the deployer address. The only thing that can go wrong with this is that chains typically need to make chain ID in transactions (EIP-155) optional for https://github.com/Arachnid/deterministic-deployment-proxy to be deployed and make it required again once they deploy that. Some chains are not aware of this, so there ends up being a period of time where CREATE2 works fine but hardhat can't deploy deterministically. If the integration is urgent, we deploy our contracts non-deterministically, yet the proxy factory deploys proxies deterministically (https://github.com/api3dao/contracts/issues/388). In such cases, without a combined proxy factory, the individual combined proxies would have to be deployed non-deterministically and it would be difficult to verify their bytecode locally.

contracts/api3-server-v1/proxies/CompositeApi3ReaderProxyV1.sol
Siegrift: I would guess that depends on the purpose of the breaking down. I think here it's justified.

contracts/api3-server-v1/proxies/InverseApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,132 @@
...
+    function read()
+        public
+        view
+        override
+        returns (int224 value, uint32 timestamp)
+    {
+        (int224 baseValue, uint32 baseTimestamp) = IApi3ReaderProxy(proxy)
+            .read();
+
+        if (baseValue == 0) {
```
Siegrift, line 39: Is it worth to check this if we're not preventing the overflow on the next line?

contracts/api3-server-v1/proxies/InverseApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,132 @@
...
+import "./interfaces/IInverseApi3ReaderProxyV1.sol";
+
+/// @title An immutable proxy contract that inverts the value returned by an
+/// IApi3ReaderProxy data feed
+/// @dev This contract implements the AggregatorV2V3Interface to be compatible
+/// with Chainlink aggregators. This allows the contract to be used as a drop-in
+/// replacement for Chainlink aggregators in existing dApps.
+/// Refer to https://github.com/api3dao/migrate-from-chainlink-to-api3 for more
+/// information about the Chainlink interface implementation.
+contract InverseApi3ReaderProxyV1 is IInverseApi3ReaderProxyV1 {
```
Siegrift, line 13: So if we ever want `A / B`, we'll do `A * (1/B)`?

Siegrift (commented):
I think this works quite well. The `InverseApi3ReaderProxyV1`, `NormalizedApi3ReaderProxyV1` and `ScaledApi3FeedProxyV1` are used rarely, but can be nicely used when needed. The most common use case will be to just multiple two Api3 proxies which the `ProductApi3ReaderProxyV1` does well.

Note about the completeness of this "combinators library". A significant unsupported use case is dApps reading values from other on-chain contracts (often done for the exchange rates). I think for such use cases it makes sense to write a specialized contract.

contracts/api3-server-v1/proxies/InverseApi3ReaderProxyV1.sol
acenolaza: Yeah, to be honest I was quite hesitant to adding this check since a zero value for a data feed is unlikely and probably wrong

acenolaza: I like the nice UX of getting and error back on cases when this happens and the caller also gets a refund of the unused gas but for our use case it might even make more sense to remove the check and wrap the math operation in `unchecked` to reduce gas even more. If we do this, we might consider documenting this so it doesn't get flagged in audits 🤷🏻

acenolaza:
> Note about the completeness of this "combinators library". A significant unsupported use case is dApps reading values from other on-chain contracts (often done for the exchange rates). I think for such use cases it makes sense to write a specialized contract.

Right, there are cases like iSEI where there are using a [composite proxy](https://seitrace.com/address/0x3e10e6eF3c602885a9Cafd9785BD2143c52fbFF1?chain=pacific-1&contract=read_contract&slug=&tab=contract) that uses an [Api3ReaderProxyV1](https://seitrace.com/address/0x7D43674e4367a74C5e57d70d34700713B9CF53F3?chain=pacific-1&contract=read_proxy&slug=&tab=contract) for reading the SEI/USD data feed rate and a specialized [API3ISeiCosmReader](https://seitrace.com/address/0xaFb041c4268Efbf217ACEea6D795FA16A3b31A2D?chain=pacific-1&tab=contract) contract that fetches the ISEI rate using CosmWasm.

There might still be one more proxy we could consider which is the [API3CompositeWrapperWithThresholding](https://fraxscan.com/address/0x66496537ECa3f784AeC62FB3B896c0086f3b1B26#code) used by dTrinity on Fraxtal or just a proxy with thresholding.

Siegrift:
> There might still be one more proxy we could consider which is the [API3CompositeWrapperWithThresholding](https://fraxscan.com/address/0x66496537ECa3f784AeC62FB3B896c0086f3b1B26#code) used by dTrinity on Fraxtal or just a proxy with thresholding.

We don't actively recommend these use cases and the dApps may have various logic here. Peg the asset, apply a cap, use a fallback oracle. The use cases are too varied here. If we ever face a dApp that requires us to help them integrate, we can help address this on a case-by-case basis.

contracts/api3-server-v1/proxies/Api3ReaderProxyV1.sol
Siegrift, line 1: I'd not touch the existing reader proxies implementation, but we've agreed to move this to a separate repo so this is no longer relevant.

contracts/api3-server-v1/proxies/InverseApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,56 @@
...
+        if (proxy_ == address(0)) {
+            revert ZeroProxyAddress();
+        }
+        proxy = proxy_;
+    }
+
+    /// @notice Returns the inverted value of the underlying IApi3ReaderProxy
+    /// @dev This inverts the 18-decimal fixed-point value using 1e36 / value.
+    /// It is assumed that operation might overflow if `baseValue` is too small
+    /// and will revert on division by zero. No safety checks are performed to
```
Siegrift, line 32: I'm thinking it would make more sense to revert after all. It seems better to not report a price rather then report an invalid one. The former is easier for consumers to protect against and they may have better protections in place.

Siegrift (approved):
Left a few comments, but these are all minor. I think once you resolve those, we can close this PR and recreate it in the other repo.

contracts/api3-server-v1/proxies/NormalizedApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,63 @@
...
+    uint8 internal immutable feedDecimals;
+
+    /// @param feed_ The address of the Chainlink AggregatorV2V3Interface feed
+    constructor(address feed_) {
+        if (feed_ == address(0)) {
+            revert ZeroProxyAddress();
+        }
+
+        uint8 feedDecimals_ = AggregatorV2V3Interface(feed_).decimals();
+        if (feedDecimals_ == 0 || feedDecimals_ > 36) {
```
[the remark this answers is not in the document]
acenolaza: The check is executed only once during contract deployment and would prevent setting up an invalid or uninitialized Chainlink feed if `feedDecimals_` is zero. Now regarding making sure is less that 36, I guess that since most frequently the `feedDecimals_` would either be 8 (stables) or 18 (others) so 36 seemed like a large enough number of decimals that would keep the scaling math safe from overflowing

contracts/api3-server-v1/proxies/adapters/interfaces/IApi3AggregatorAdapter.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.27;
+
+import "../../../../vendor/@chainlink/contracts@1.2.0/src/v0.8/shared/interfaces/AggregatorV2V3Interface.sol";
+import "../../../../interfaces/IApi3ReaderProxy.sol";
+
+interface IApi3AggregatorAdapter is IApi3ReaderProxy, AggregatorV2V3Interface {
```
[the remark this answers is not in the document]
acenolaza: Maybe the name of the interface is not clear but idea was to have a single interface for Api3/Chainlink adapter proxies. It also helps a bit with DRY by not having to re-define the `FunctionIsNotSupported` error

contracts/api3-server-v1/proxies/adapters/BaseApi3AggregatorAdapter.sol
```diff
@@ -0,0 +1,102 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.27;
+
+import "./interfaces/IApi3AggregatorAdapter.sol";
+
+/// @title Base implementation of AggregatorV2V3Interface for API3 proxies
```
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/a7318be313a8ef282ff3c890bd4a07e796f2e5bf

contracts/api3-server-v1/proxies/ProductApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,62 @@
...
+            revert SameProxyAddress();
+        }
+        proxy1 = proxy1_;
+        proxy2 = proxy2_;
+    }
+
+    /// @notice Returns the current value and timestamp of the rate composition
+    /// between two IApi3ReaderProxy proxies by multiplying their values
+    /// @dev There is a risk of multiplication overflowing if result is not
+    /// suitable for int256 type. The timestamp is the current block timestamp
```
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/23a40ed46a96bbc94b55c32c4fce86e084e90981

contracts/api3-server-v1/proxies/NormalizedApi3ReaderProxyV1.sol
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/396584612eacfbc7faeda847da8cde9f2c92d1eb

contracts/api3-server-v1/proxies/NormalizedApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,63 @@
...
+            revert UnsupportedFeedDecimals();
+        }
+        feed = feed_;
+        feedDecimals = feedDecimals_;
+    }
+
+    /// @notice Returns the normalized to 18 decimals value from the latest value
```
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/396584612eacfbc7faeda847da8cde9f2c92d1eb

contracts/api3-server-v1/proxies/NormalizedApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,63 @@
...
+    /// of underlying Chainlink feed
+    /// @return value The normalized signed fixed-point value (18 decimals)
```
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/396584612eacfbc7faeda847da8cde9f2c92d1eb

contracts/api3-server-v1/proxies/InverseApi3ReaderProxyV1.sol
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/398d1f3bdf2fd82775fd437165ea4a79b02560f7

contracts/api3-server-v1/proxies/InverseApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,56 @@
...
+    /// optimize gas usage. Use with trusted feeds only.
```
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/ff6126e8c22711733e53d5a33d94a1d8a685fe04

contracts/api3-server-v1/proxies/adapters/ScaledApi3FeedProxyV1.sol
```diff
@@ -0,0 +1,137 @@
...
+/// @dev This contract assumes the source proxy always returns values with
+/// 18 decimals (as all IApi3ReaderProxy-compatible proxies do)
+contract ScaledApi3FeedProxyV1 is IScaledApi3FeedProxyV1 {
+    using ProxyUtils for int256;
+
+    /// @notice IApi3ReaderProxy contract address
+    address public immutable override proxy;
+
+    /// @notice Decimals used to scale the IApi3ReaderProxy value
+    uint8 public immutable override targetDecimals;
```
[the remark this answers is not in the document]
acenolaza: I thought that it might be semantically better for callers to have a separate `public` function so they could either use:
* `AggregatorV2V3Interface(proxy).decimals()`
* `IScaledApi3FeedProxyV1(proxy).targetDecimals()`
but I don't have a strong case to defend this so I'll just make it `private`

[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/b5aa03c564068f4ad2b1d08e163766d251e98011

contracts/api3-server-v1/proxies/Api3ReaderProxyV1.sol
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/9b04883638a4e7b3f87424908362fe5917b0a922

contracts/api3-server-v1/proxies/adapters/interfaces/IApi3AggregatorAdapter.sol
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/8560884d55f99656d379e7f78c00cb7ccd7c494f

contracts/api3-server-v1/proxies/adapters/BaseApi3AggregatorAdapter.sol
```diff
@@ -0,0 +1,102 @@
...
+/// @notice This abstract contract provides common implementation of
```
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/contracts/pull/472/commits/8560884d55f99656d379e7f78c00cb7ccd7c494f

contracts/api3-server-v1/proxies/NormalizedApi3ReaderProxyV1.sol
```diff
@@ -0,0 +1,63 @@
...
+/// @dev This contract implements the AggregatorV2V3Interface to be compatible
+/// with Chainlink aggregators. This allows the contract to be used as a drop-in
+/// replacement for Chainlink aggregators in existing dApps.
+/// Refer to https://github.com/api3dao/migrate-from-chainlink-to-api3 for more
+/// information about the Chainlink interface implementation.
+contract NormalizedApi3ReaderProxyV1 is
+    BaseApi3AggregatorAdapter,
+    INormalizedApi3ReaderProxyV1
+{
+    using ProxyUtils for int256;
```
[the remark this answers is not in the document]
acenolaza: This function will get embedded in deployed contracts bytecode but I still think it helps with keeping the code DRY, easier to maintain and readability

PR #492, opened 2025-05-14
Updates RPC provider configurations

bdrhn9:
Coverage check for RPC providers has been conducted. Any unreliable providers have been either removed or replaced with more stable alternatives.


data/chains/x-layer-sepolia-testnet.json
```diff
@@ -15,6 +15,10 @@
   "providers": [
     {
       "alias": "default",
+      "rpcUrl": "https://rpc.ankr.com/xlayer_testnet/"
```
hiletmis, line 18: 👍

PR #493, opened 2025-05-14
Add Morpho, tunnl and TakoTako to dApp details

Siegrift:


data/dapps/morpho.json
```diff
@@ -0,0 +1,15 @@
+{
+  "aliases": {
+    "morpho-mvl-usdc-770-lltv": {
```
Siegrift, line 3: The title follows the naming [on the Morpho website](https://app.morpho.org/ethereum/market/0xb323495f7e4148be5643a4ea4a8221eef163e4bccfdedc2a6f4696baacbc86cc/wsteth-usdc). Under the hood, the market is identified by [5 immutable parameters](https://docs.morpho.org/overview/concepts/market/#the-five-parameters):
1. Collateral asset = required
2. Loan asset = required
3. LLTV = required. The [DAO approved values](https://docs.morpho.org/overview/concepts/market/#lltvs) can be written in 3 digits, so I opted to percentage multiplied by 10. 
4. oracle address = in practice this would be more like "oracle type", but it's implied that oracle is Api3 so can be omitted
5. IRM = There is [only one DAO approved one](https://docs.morpho.org/overview/concepts/market/#irm) so we can omit it as well. In the future if there are more, we can just add it as suffix to the alias.

We obviously also need the "morpho" somewhere in the dApp title. So altogether, I chose the following convention: `morpho-<COLLATERAL>-<BORROW>-<LLTV_E1>-lltv`. The title follows similar pattern, but token names are used and the percentage is in human readable form. The description follows similar pattern.

data/dapps/tunnl.json
```diff
@@ -0,0 +1,6 @@
+{
+  "aliases": {
+    "tunnl": { "chains": ["polygon"], "title": "tunnl" }
```
Siegrift, line 3: These are not on Defillama, but according to their homepage I assume `tunnl` (lowercased title) is how they want to be branded.

data/dapps/morpho.json
Siegrift: Yeah, although any decision here is arbitrary so I'd stick with this one. I don't have any hopes this convention will outlive us, but it does the job.

PR #515, opened 2025-06-04
Migrate core-testnet to new implementation

hiletmis:
Related to https://github.com/api3dao/data-feeds/issues/3019

Also see https://docs.coredao.org/docs/Dev-Guide/network-config


bdrhn9 (approved):
Looks good.

PR #527, opened 2025-06-18
Remove unused dApp aliases

Siegrift:
See: https://api3workspace.slack.com/archives/C05AGC6RBMK/p1750154031226489


.changeset/sweet-tires-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/contracts': major
```
Siegrift, line 2: Even though they are unused, I decided to do a major version anyway. There are some Api3 projects for which this will cause breakage.

PR #529, opened 2025-06-18
Updates RPC provider configurations

bdrhn9:
Coverage/reliability check for RPC providers has been conducted. Any unreliable providers have been either removed or replaced with more stable alternatives.


hiletmis (approved):
LGTM

Thanks for updating unreliable RPCs. Especially `ethereum-holesky-testnet` was problematic.

PR #533, opened 2025-06-23
Csur-2025-06-23

hiletmis:
Related to https://github.com/api3dao/data-feeds/issues/3095


data/chains/katana.json
```diff
@@ -0,0 +1,27 @@
...
+      },
+      "url": "https://explorer.katanarpc.com/api"
+    },
+    "browserUrl": "https://explorer.katanarpc.com/"
+  },
+  "id": "747474",
+  "name": "Katana",
+  "providers": [
+    {
+      "alias": "default",
```
bdrhn9, line 17: I was about to suggesting to add https://rpc.katanarpc.com introduced in https://github.com/ToniRamirezM/chains/blob/389f820c68fb361ecc8b5637d1050606fdcf619a/_data/chains/eip155-747474.json#L4 but checked that it isn't working. So it's okay to go with these providers, maybe we can consider to run node for the chain.

bdrhn9 (approved):
LGTM.

PR #561, opened 2025-07-03
Adds compute-dapp-id CLI command

### LINKED ISSUE #420: Have a CLI command to derive dApp ID
Given an dApp alias and chain ID. This is mainly beneficial for searchers, who currently have to clone the repo to use the function.

acenolaza:
Closes #420 

### Implementation details

* Allows the user to compute dApp IDs for unknown dApp aliases and/or chain IDs
* On strict mode `compute-dapp-id` will exit for unknown dApp aliases/chain IDs

[Documentation](https://docs.api3.org/oev-searchers/in-depth/data-feeds/#dapp-ids) might need to be updated

### DIFF
--- a/.changeset/every-breads-knock.md
+++ b/.changeset/every-breads-knock.md
@@ -1,0 +1,5 @@
+---
+'@api3/contracts': minor
+---
+
+Added the `compute-dapp-id` CLI command


src/cli.ts
```diff
@@ -132,4 +133,43 @@ yargs(hideBin(process.argv))
...
+        if (args.strict) {
+          console.error(message);
+          process.exit(1);
+        }
+        console.warn(message);
+      }
+
+      console.log(`dApp alias: ${args['dapp-alias']}\nchain: ${chain?.name ?? args['chain-id']}`);
+
+      if (dappInfo) {
```
Siegrift, line 166: We should also verify that the chain alias is present for the dApp.

src/cli.ts
```diff
@@ -132,4 +133,43 @@ yargs(hideBin(process.argv))
...
+        const [, aliasInfo] = Object.entries(dappInfo.aliases).find(([alias]) => alias === args['dapp-alias'])!;
+        if (aliasInfo.description) console.log(`\nℹ️  ${aliasInfo.description}`);
+      }
+
+      const dappId = unsafeComputeDappId(args['dapp-alias'], args['chain-id']);
+      console.log(`\ndApp ID: ${dappId.toString()}`);
```
Siegrift, line 172: Maybe:
```suggestion
      console.log(`\n• dApp ID: ${dappId.toString()}`);
```

Siegrift (changes requested):

src/cli.ts
```diff
@@ -132,4 +133,43 @@ yargs(hideBin(process.argv))
       );
     }
   )
+  .command(
+    'compute-dapp-id',
+    'Computes the dApp ID for a given dApp alias and chain ID',
+    {
+      'dapp-alias': dappAlias,
+      'chain-id': chainId,
```
bbenligiray, line 141: Is this command public-facing or internal-facing? The rule is internal-facing interfaces use chain alias (because it's convenient and readable) and public-facing interfaces use chain ID (because the public don't know the aliases)

bbenligiray:
(Needs a changeset)

src/cli.ts
Siegrift: Public. I also want to document it [here](https://docs.api3.org/oev-searchers/in-depth/data-feeds/#dapp-ids).

Siegrift (commented):
> Not that I mind it but is this suddenly needed?

It's not, but there are times where this is handy. We often double check that that the dApp ID of the dApp proxies is correct (among others uses). I saw it in the backlog while cleaning up issues and decided it's time to get this over with.

src/cli.ts
Siegrift: I think the `unsafeComputeDappId` is sufficient for that purpose. The CLI is there only so that you don't have to bother with creating any script if you just want to quickly verify that.

src/cli.ts
acenolaza: https://github.com/api3dao/contracts/pull/561/commits/8ffffe51d9a2f1db31326da7c407ccc490c761ac

Siegrift (approved):
👍 LGTM

@acenolaza Can you also document it in the docs as I mentioned [here](https://github.com/api3dao/contracts/pull/561/files#r2184600088). Just a brief way on how to execute it, similarly how it's done [here](https://docs.api3.org/dapps/integration/contract-integration.html#printing-api3readerproxyv1-addresses)

Issue #363, opened 2025-02-04
Test the CLI command


bbenligiray:
Assuming that the CLI command was going to be very static, I've been just testing all conditions manually. We do appear to be [adjusting](https://github.com/api3dao/contracts/pull/360) it though, so automated tests would be nice. I would refer to [airnode-admin e2e tests](https://github.com/api3dao/airnode/tree/master/packages/airnode-admin/test/e2e) for how that could be done.

hiletmis:
The CLI provides two commands:

- **`print-api3readerproxyv1-address`**
- **`compute-dapp-id`**

**Testing `compute-dapp-id`**
The `compute-dapp-id` command can be tested with deterministic inputs using an iteration. This makes it straightforward to validate expected outputs without external dependencies.

**Testing `print-api3readerproxyv1-address`**
This command depends on a real deployed proxy contract to return consistent results. For testing, there are two main approaches:

1. **Mock**
   Simulate the call:
   ```ts
   code = await provider.getCode(proxyAddress); 
   ```
2. **Use a Real Proxy**
Provide a known deployed proxy address on a specific network for an integration-style test.

Which strategy should be followed?

bbenligiray:
Actual RPC provider calls in an e2e test is acceptable but testing this will require a lot of calls if different aliases/chain IDs/dAPI names are tried. So I would go with mocking the call.

PR #572, opened 2025-08-04
Update dependencies to latest

hiletmis:
Closes https://github.com/api3dao/contracts/issues/573

~Regarding https://github.com/api3dao/contracts/issues/363 CLI command has been updated w/o e2e tests.~ Reverted

~zod migration is being tracked via https://github.com/api3dao/contracts/pull/575~ Synced


bbenligiray:
Any modification to the CLI tool is blocked by the implementation of e2e tests

package.json
```diff
@@ -70,43 +70,43 @@
...
+    "@nomicfoundation/hardhat-verify": "^2.1.1",
     "@openzeppelin/merkle-tree": "^1.0.8",
     "@typechain/ethers-v6": "^0.5.1",
     "@typechain/hardhat": "^9.1.0",
     "@types/chai": "^4.3.20",
-    "@types/jest": "^29.5.14",
+    "@types/jest": "^30.0.0",
     "@types/mocha": "^10.0.10",
-    "@types/node": "^20.19.0",
+    "@types/node": "^24.2.0",
```
bdrhn9, line 84: According to [this](https://stackoverflow.com/a/52404327), major and minor version should be same with node version, which is 22.18.x.

test-extended/api3-server-v1/Api3MarketV2.sol.ts
```diff
@@ -3,10 +3,12 @@ import type { HardhatEthersSigner } from '@nomicfoundation/hardhat-ethers/signer
 import * as helpers from '@nomicfoundation/hardhat-network-helpers';
 import { expect } from 'chai';
 import type { BytesLike, HDNodeWallet } from 'ethers';
-import { ethers } from 'hardhat';
+import hardhat from 'hardhat';
```
bdrhn9, line 6: Do we need change on this line?

package.json
hiletmis: This is a bit controversial. But it should be better to keep node and @types/node sycned

test-extended/api3-server-v1/Api3MarketV2.sol.ts
hiletmis: Tests fail due to mixed usage of common js and esm. It is inevitable as hardhat is common js

```
import { ethers } from 'hardhat';
         ^^^^^^
SyntaxError: Named export 'ethers' not found. The requested module 'hardhat' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'hardhat';
const { ethers } = pkg;
```

hiletmis: Your node version must be v20

bdrhn9:
I'm generally okay with the PR but I wouldn't prefer to bump `pnpm` and `node` versions.

hiletmis:
In fact, here’s [Hardhat v3](https://hardhat.org/migrate-from-hardhat2). It now supports ESM, but I think it’s still a bit early to migrate—it’s only a day old. 

It also requires Node 22.10 or later, so finalizing this PR at the current level should save us from having to deal with multiple major migrations at once.

bbenligiray (approved):

Issue #574, opened 2025-08-07
Migrate zod to v4


hiletmis:
A new major version update of `zod` has been released recently.

[Reference](https://zod.dev/v4/changelog) for migration.

See https://api3workspace.slack.com/archives/C01N6001YCE/p1754406409097589

PR #575, opened 2025-08-07
Migrate zod to v4

hiletmis:
Closes https://github.com/api3dao/contracts/issues/574

This PR upgrades the zod package to version 4.

In addition to the migration, missing test cases have been added to ensure full coverage and validation of the updated schema logic.


src/types.test.ts
```diff
@@ -128,3 +123,122 @@ describe('dappSchema', () => {
...
+        {
+          code: 'custom',
+          path: ['providers', 'alias'],
+          message: "cannot contain duplicate 'alias' values",
+        },
+      ])
+    );
+  });
+
+  it('should reject if "default" or "public" provider does not have rpcUrl', () => {
```
bdrhn9, line 232: Can you also add provider with alias `public` (with `homepageUrl` attribute) into `invalidProviders` array?

src/types.ts
```diff
@@ -110,10 +115,24 @@ export const aliasSchema = z.string().regex(/^[\da-z-]+$/);
...
+  }
+
+  const isKnown = CHAINS.some((chain) => chain.alias === value);
+  if (!isKnown) {
+    ctx.addIssue({
+      code: 'custom',
+      message: `Invalid chain alias: ${value}`,
+    });
+  }
+});
```
bdrhn9, line 135:
```suggestion
export const chainAlias = aliasSchema.refine((value) => CHAINS.some((chain) => chain.alias === value), {
  error: (iss) => `Invalid chain alias: ${iss.input}`,
});
```

I'd stick to original implementation in the v4 way.

src/types.ts
```diff
@@ -10,49 +10,54 @@ export const chainExplorerAPIKeySchema = z.object({
...
 
 export const chainProvidersSchema = z.array(chainProviderSchema).superRefine((providers, ctx) => {
   if (!providers.some((p) => p.alias === 'default')) {
-    ctx.addIssue({
-      code: z.ZodIssueCode.custom,
+    ctx.issues.push({
+      code: 'custom',
       path: ['providers', 'alias'],
       message: "a provider with alias 'default' is required",
+      input: providers.map((p) => p.alias),
```
bdrhn9, line 41: Using the whole `providers` object is another option, but I'd also prefer using only the mapped relevant values 👍🏼.

That being said, since this issue is at the array level and we’re passing the whole array as `input`, I’d leave the `path` attribute empty `[]`. Basically, my main approach is: if it’s an array-level error, leave the path empty; if it’s an error for a specific element in the array, set the `path` precisely to that element.

src/types.ts
```diff
@@ -10,49 +10,54 @@ export const chainExplorerAPIKeySchema = z.object({
...
 
   providers.forEach((p) => {
     if ((p.alias === 'default' || p.alias === 'public') && !p.rpcUrl) {
-      ctx.addIssue({
-        code: z.ZodIssueCode.custom,
+      ctx.issues.push({
+        code: 'custom',
         path: ['providers', 'rpcUrl'],
         message: "providers with alias 'default' or 'public' must also have an 'rpcUrl'",
+        input: p.rpcUrl,
```
bdrhn9, line 60:
```suggestion
        path: [index, 'rpcUrl'],
        message: "providers with alias 'default' or 'public' must also have an 'rpcUrl'",
        input: p.rpcUrl,
```

If you'd like to follow my approach mentioned above.

.changeset/tough-cups-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/contracts': major
```
bdrhn9, line 2: 👍🏼

src/types.ts
hiletmis: I have tried that at first but tests were failing due to a bug in zod:

```
    expect(received).toThrow(expected)

    - Expected message  - 0
    + Received message  + 1

      [
        {
    +     "origin": "string",
          "code": "invalid_format",
          "format": "regex",
          "pattern": "/^[\\da-z-]+$/",
          "path": [],
          "message": "Invalid string: must match pattern /^[\\da-z-]+$/"
        },
        {
          "code": "custom",
          "path": [],
          "message": "Invalid chain alias: Mantle"
        }
      ]
 ```
 
 I have added `origin` but then I got this error:
```
Object literal may only specify known properties, and 'origin' does not exist in type '$ZodIssueInvalidStringFormat'.ts(2353)
```

Which is true and can be seen [here](https://github.com/colinhacks/zod/blob/592b7b5553b5ef64474ec4ff6d4a9a5fc3c9d409/packages/zod/src/v4/core/errors.ts#L46). But `origin` is being added anyway like following [ZodCheckStringFormat](https://github.com/colinhacks/zod/blob/592b7b5553b5ef64474ec4ff6d4a9a5fc3c9d409/packages/zod/src/v4/core/checks.ts#L808)

There are some issues related to regex on zod repo but I couldn't find any about this specific situation. But you can see the actual behavior is too emit `origin` here is an [issue](https://github.com/colinhacks/zod/issues/4964).

src/types.ts
hiletmis: It is an array error indeed, leaving it empty might be preferable

bdrhn9 (approved):

PR #581, opened 2025-08-18
Add tests for the CLI command

hiletmis:
Related to https://github.com/api3dao/contracts/issues/363

**There are some limitations testing the CLI command**
The child process spawned by execSync:
* Does not inherit the Jest mocks
* Loads its own modules (ethers, CLI methods) from scratch
* Creates real JsonRpcProvider and contract objects inside the child process
* execSync is synchronous, but many ethers calls are async (await provider.getCode(), readDataFeedWithDapiNameHash)

to test with mocks the code should be refactored so that core logic is a function and can be exported.

### DIFF
--- a/src/cli.test.ts
+++ b/src/cli.test.ts
@@ -8,8 +8,7 @@ type CommandArg = [string, string | number | boolean];
 
 describe('cli tests', () => {
   const execCommand = (command: string, ...args: CommandArg[]) => {
-    const quote = (val: string) => `"${val}"`;
-    const formattedCommand = `${command} ${args.map(([c, a]) => `${c} ${quote(String(a))}`).join(' ')}`;
+    const formattedCommand = [command, ...args.flatMap(([c, a]) => [c, String(a)])].join(' ');
     const goExecSync = goSync(() =>
       execSync(`node ${CLI_EXECUTABLE} ${formattedCommand}`, {
         stdio: ['pipe', 'pipe', 'pipe'],


src/cli.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+type CommandArg = [string, string | number | boolean];
+
+describe('cli tests', () => {
+  const execCommand = (command: string, ...args: CommandArg[]) => {
+    const quote = (val: string) => `"${val}"`;
+    const formattedArgs = args
+      .map(([c, a]) => {
+        // if args is array then quote each elem and separate them with space
+        if (Array.isArray(a)) return `${c} ${a.map((element) => quote(element)).join(' ')}`;
```
bbenligiray, line 15: The `CommandArg` type definition above says `a`'s type can be `string | number | boolean`. I don't see why we're handling the case where `a` is an array here.

src/cli.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+import { goSync } from '@api3/promise-utils';
+
+const CLI_EXECUTABLE = 'dist/src/cli.js';
+
+type CommandArg = [string, string | number | boolean];
+
+describe('cli tests', () => {
+  const execCommand = (command: string, ...args: CommandArg[]) => {
+    const quote = (val: string) => `"${val}"`;
```
bbenligiray, line 11: I think you should also get rid of `quote()` considering that we're not recommending the user to enclose argument values in quotation marks anywhere. That was an airnode-admin thing where seed phrases needed to be enclosed because they have spaces in them.

bbenligiray (approved):
lgtm

Issue #594, opened 2025-09-03
Somnia is in skipped native currency validation list - remove when SOMI listed and Cint


hiletmis:
* SOMI has just started trading and not listed
* `somnia` is in survey

When SOMI is listed and somnia integration is done it will be removed from skipped native currency validation list

PR #595, opened 2025-09-03
Csur-2025-09-03

hiletmis:
Related to https://github.com/api3dao/data-feeds/issues/3375 and https://github.com/api3dao/contracts/issues/594


bdrhn9 (approved):

bdrhn9:
> Why did validate-verify print `Error: somnia OevAuctionHouse native currency rate proxy address is 0x0000000000000000000000000000000000000000 while it should have been 0xe3c6253B057f5447a696f9e5293Cc1F98f66aA3a` even though the alias was added to the list? I couldn't figure it out.

I think it has always been like this (see [here](https://github.com/api3dao/contracts/actions/runs/15825509596/job/44604548305)), except for chains with ETH as the native currency. These validations will be resolved after the manager-multisig steps are completed (which sets the AirseekerRegistry address and OEV network proxies). If this feels like a problem, we can add a new constant, `skippedChainAliasesForPendingManagerMultisig`, that ignores chains until the manager-multisig steps are done. We can also add this constant setting step to the Csur checklist.

Issue #547, opened 2025-06-27
Contract verification migration to Etherscan v2


hiletmis:
Contract verification prints following warning:

`[WARNING] Network and explorer-specific api keys are deprecated in favour of the new Etherscan v2 api. Support for v1 is expected to end by May 31st, 2025. To migrate, please specify a single Etherscan.io api key the apiKey config value.`

See [info](https://info.etherscan.com/switch-to-etherscan-api-v2-by-may-31-2025/)

Siegrift:
It would be nice to make this work. I'm currently spending quite a bit of time making this work for the chains I deploy proxies or adapters for. For example these are the changes (of hardhat config) I do to make it work for Katana:
```
  etherscan: {
    ...hardhatConfig.etherscan(),
    apiKey: {
      katana: process.env.ETHERSCAN_API_KEY_KATANA!,
    },
    customChains: [
      {
        network: 'katana',
        chainId: 747_474,
        urls: {
          apiURL: 'https://api.etherscan.io/v2/api?chainid=747474',
          browserURL: 'https://katanascan.com/',
        },
      },
    ],
  },
  networks: {
    ...hardhatConfig.networks(),
    katana: {
      url: 'https://rpc.katana.network',
      chainId: 747_474,
      accounts: { mnemonic: process.env.MNEMONIC ?? '' },
    },
  },
```

I then supply the MNEMONIC and the ETHERSCAN_API_KEY_KATANA when executing the deployment command. Note that the verification needs to happen explicitly by the script (e.g. [here](https://github.com/api3dao/data-feed-proxy-combinators/blob/f26029757662858906bb5b78a49a7c5d9940c50b/deploy/003_deploy_ProductApi3ReaderProxyV1.ts#L118)) and I am not sure if there is a way to preserve the old verification mechanism.

PR #609, opened 2025-09-16
Migrate contract verification to Etherscan V2 API

hiletmis:
Closes https://github.com/api3dao/contracts/issues/547

## Acknowledgements
* Etherscan V1 is deprecated and V2 is encouraged by hardhat
* V2 no longer needs an API url specified on network object but it is still needed on [hardhat config](https://github.com/NomicFoundation/hardhat/blob/0a145098cd1f50f6b46fc17ae9a19d393c56d9d3/packages/hardhat-verify/src/types.ts#L1)
* V2 uses one API key for all [supported chains](https://docs.etherscan.io/supported-chains)
* Default `verify:verify` action calls Etherscan API although we have a few Blockscout and other explorers. 

## Proposed changes
* An API key for every chain usage is dropped and single `ETHERSCAN_API_KEY` is introduced.
* Redundant API section is removed from chain objects
* Added Blockscout and Sourcify support
* A new field `provider` is added to chain object to allow verify process to choose between APIs (Blockscout, Etherscan, Sourcify)
* It is a bit controversial but I marked these changes as `major` as `HardhatEtherscanConfig` is [modified](https://github.com/api3dao/contracts/blob/33d9d60cfef34085d65d6b289a23f7b6f1d35983/src/types.ts#L112)

## What is missing?
* Chains that use "other" block explorers can also be verified using Blockscout scheme but since Blockscout is not accepting an API key as dummy key is already [hardcoded](https://github.com/NomicFoundation/hardhat/blob/0a145098cd1f50f6b46fc17ae9a19d393c56d9d3/packages/hardhat-verify/src/internal/blockscout.ts#L26) we can not specify any API key (if needed). So far this is not a big deal and we can check this issue on hardhat v3 migration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,7 +359,8 @@ console.log(hardhatConfig.getEnvVariableNames());
 /*
 [
   'MNEMONIC',
-  'ETHERSCAN_API_KEY'
+  'ETHERSCAN_API_KEY',
+  'HARDHAT_HTTP_RPC_URL_APECHAIN_ARBITRUM_SEPOLIA_TESTNET',
   ...
 ]
 */


.changeset/slick-words-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@api3/contracts': major
```
bbenligiray, line 2: `@api3/chains` used to export `hardhatConfig` for this repo to use. Afaik no `@api3/contracts` user uses `hardhatConfig` nor do we intend them to (unlike `viemConfig` for example), so while we're doing a major release, we can just remove the `hardhatConfig` export so that future changes to this don't require major releases.

src/types.ts
```diff
@@ -13,9 +13,12 @@ export const chainExplorerAPISchema = z.object({
   url: z.url(),
 });
 
+export const chainExporerProviderSchema = z.enum(['etherscan', 'blockscout', 'sourcify', 'other']);
```
bbenligiray, line 16: Typo in the variable name

bbenligiray, line 16: Sourcify is not a block explorer though, right?

.changeset/slick-words-remain.md
bbenligiray: That being said, the problem here is not only `hardhatConfig` is changing but also `CHAINS` is changing.

hiletmis: There is [not much](https://github.com/search?q=org%3Aapi3dao%20import%20%7B%20hardhatConfig%20%7D%20from%20%27%40api3%2Fcontracts%27%3B&type=code). But I am sure they can take care of themselves.

src/types.ts
hiletmis: Also there is `blockExplorerProvider` on verify script. That is also misleading. How about using `verificationProviderSchema` for types and `verificationProvider` variable.

bbenligiray: `verificationApi.key` appears to be a part of the chain schema, I guess that's a mistake.

bbenligiray: I would do `verificationApi.type` rather than `verificationApi.provider` because we already have a `providers` entry that's totally unrelated.

deploy/3_verify.ts
```diff
@@ -10,31 +11,35 @@ module.exports = async () => {
     throw new Error(`${network.name} is not supported`);
   }
 
+  const blockExplorerProvider = CHAINS.find((chain) => chain.alias === network.name)?.explorer?.provider;
+  const verifyTask =
+    blockExplorerProvider === 'other' ? 'verify:blockscout' : `verify:${blockExplorerProvider ?? 'verify'}`;
```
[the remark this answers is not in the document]
hiletmis: This allows 'other' to be processed as blockscout. `verify:${blockExplorerProvider ?? 'verify'}` this part covers following:
`verify:etherscan`
`verify:blockscout`
`verify:sourcify`

and if not defined:
`verify:verify`

bdrhn9 (commented):
In README.md, `### hardhatConfig.etherscan()`, `### hardhatConfig.getEnvVariableNames()` fields need to be updated. Additionally, we can add recently introduced hardhatConfig.blockscout().

After these minor things are addressed, I'm happy with the implementation. Now, chain schema looks cleaner.

src/hardhat-config.ts
```diff
@@ -31,37 +34,41 @@ export function etherscan(): HardhatEtherscanConfig {
...
+  return {
+    enabled: true,
+    customChains: CHAINS.filter(
+      (chain) => chain.verificationApi?.type === 'blockscout' || chain.verificationApi?.type === 'other'
+    ).map((chain) => ({
+      network: chain.alias,
+      chainId: Number(chain.id),
+      urls: {
+        apiURL:
+          chain.verificationApi?.type === 'blockscout' || chain.verificationApi?.type === 'other'
```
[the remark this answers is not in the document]
hiletmis: It is because a typescript error:
```
Property 'url' does not exist on type '{ type: "etherscan"; } | { type: "sourcify"; } | { type: "blockscout"; url: string; } | { type: "other"; url: string; }'.
  Property 'url' does not exist on type '{ type: "etherscan"; }'.ts(2339)
```

deploy/3_verify.ts
```diff
@@ -10,31 +11,35 @@ module.exports = async () => {
     throw new Error(`${network.name} is not supported`);
   }
 
+  const verificationApiType = CHAINS.find((chain) => chain.alias === network.name)?.verificationApi?.type;
+  const verifyTask =
+    verificationApiType === 'other' ? 'verify:blockscout' : `verify:${verificationApiType ?? 'verify'}`;
```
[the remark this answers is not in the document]
hiletmis: Actual usage of `verify:verify` is letting `hardhat.config` to decide which API will be used by specifying only one of the three verification types.

Since we use all three verification types, the default `verify:verify` command may not behave as expected. Therefore, we explicitly specify each type and define a default action even if there is no `verificationApi` is stated.

We can change it to throw or skip verification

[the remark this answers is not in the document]
bdrhn9: > We can change it to throw or skip verification in case there is no verificationApi 

Skipping verification if no `verificationApi` provided sounds expected to me, wdyt?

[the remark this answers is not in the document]
hiletmis: I think so

bbenligiray:
I like the new schema. Please make sure that the users (api3-market in particular) update accordingly. I'm not approving because I didn't/won't properly review the rest.

Issue #610, opened 2025-09-23
Validate coverage of Etherscan V2 API


bdrhn9:
Related to https://github.com/api3dao/contracts/issues/547.

Integrate the https://api.etherscan.io/v2/chainlist to improve validation and consistency for supported chains.

**Some validations to consider:**
* Verify if a chain is still supported by checking against the chainlist API.
* Validate that the configured `blockExplorerUrl` matches the value returned by the chainlist endpoint.
* Add a warning if a chain is configured with a non-etherscan `verificationApi` type, even though the chainlist API indicates that `etherscan` is supported.

bdrhn9:
> Maybe the only question is should we always prefer Etherscan over other block explorers?

The Etherscan V2 API can be considered the default verification method this repository depends on, right? So yeah, I think we can prefer Etherscan over others if it works as expected.

PR #614, opened 2025-09-26
Verification API compliance check

hiletmis:
Closes https://github.com/api3dao/contracts/issues/610

This PR introduces a new script that can be called to check block explorers and verification APIs against Etherscan V2 supported chain API.

Also to comply with the API following changes are made:

```
Update block explorer and verification API for following chains:

- avalanche
- avalanche-testnet
- katana
- monad-testnet
- moonbeam
- sei
- sei-testnet

Update verification API for:

- polygon-zkevm-sepolia-testnet
```

**Verification script changes**

`constructorArguments` seems to be an invalid field in verify script. The actual parameter is `constructorArgsParams` according to [hardhat-verify](https://github.com/NomicFoundation/hardhat/blob/0a145098cd1f50f6b46fc17ae9a19d393c56d9d3/packages/hardhat-verify/src/index.ts#L54). `contructorArguments` is mentioned [here](https://github.com/NomicFoundation/hardhat/blob/0a145098cd1f50f6b46fc17ae9a19d393c56d9d3/packages/hardhat-verify/src/internal/etherscan.ts#L174) but this is encoded -has a typo- (also string) and it is originated [here](https://github.com/NomicFoundation/hardhat/blob/0a145098cd1f50f6b46fc17ae9a19d393c56d9d3/packages/hardhat-verify/src/internal/tasks/etherscan.ts#L160) using `constructorArgsParams`

It was working ok with API V1 w/o any problem because it meant that no constructor arguments are provided.

In V2 it looks like it is mandatory to provide constructor arguments

**Why this issue didn't come up on migration?**

- All etherscan validations were already done.
- I have deployed contracts non-deterministic to test if API v2 is working and apparently since byte code wasn't changed the API also marked them as verified and skipped actual verification.

There is a similar [issue](https://github.com/NomicFoundation/hardhat/issues/945) discussed BC 2020

### DIFF
--- a/scripts/validate-verification-api.ts
+++ b/scripts/validate-verification-api.ts
@@ -66,13 +66,13 @@ async function main() {
 
     if (chain.verificationApi?.type === 'etherscan' && chain.blockExplorerUrl !== supportedChain.blockexplorer) {
       errors.push(
-        `🔴 ${supportedChain.chainname} (ID: ${supportedChain.chainid}) has a different block explorer URL. Local: ${chain.blockExplorerUrl}, Etherscan: ${supportedChain.blockexplorer}`
+        `🔴 ${chain.alias} (ID: ${chain.id}) has a different block explorer URL. Local: ${chain.blockExplorerUrl}, Etherscan: ${supportedChain.blockexplorer}`
       );
     }
 
     if (!chain.verificationApi || chain.verificationApi.type !== 'etherscan') {
       warnings.push(
-        `⚠️ ${supportedChain.chainname} (ID: ${supportedChain.chainid}) supports Etherscan v2 API but verificationApi is not set in the local chain list.`
+        `⚠️ ${chain.alias} (ID: ${chain.id}) supports Etherscan v2 API but verificationApi is not set in the local chain list.`
       );
     }
   }


scripts/check-etherscan-api.ts
```diff
@@ -0,0 +1,92 @@
...
+const ResultSchema = z.object({
+  chainname: z.string(),
+  chainid: z.string(),
+  blockexplorer: z.string(),
+  apiurl: z.string(),
+  status: z.number(),
+  comment: z.string(),
+});
+
+const EtherscanV2SupportListSchema = z.object({
```
bdrhn9, line 15: We defined schema here but utilized it only for inferring type `EtherscanV2SupportList`. I'd suggest using this to parse response from URL.

scripts/validate-verification-api.ts
```diff
@@ -0,0 +1,93 @@
...
+    throw new Error('Etherscan API is not reachable');
+  }
+
+  const response = goResponse.data;
+
+  if (!response.ok) {
+    throw new Error(`Etherscan API responded with status: ${response.status}`);
+  }
+
+  const data = (await response.json()) as z.infer<typeof etherscanV2SupportListSchema>;
```
bdrhn9, line 39:
```suggestion
  const parseData = etherscanV2SupportListSchema.safeParse(await response.json());
  if (!parseData.success) {
    const errors = parseData.error.issues.map((issue) => `  path: '${issue.path.join('.')}' => '${issue.message}' `);
    throw new Error(`Etherscan API response schema mismatch:\n${errors.join('\n')}\n`);
  }
  const { data } = parseData;
```

Actually, I meant more like this.

scripts/validate-verification-api.ts
```diff
@@ -0,0 +1,93 @@
...
+  for (const supportedChain of etherscanSupportedChains) {
+    const chain = CHAINS.find((c: Chain) => c.id === supportedChain.chainid);
+
+    if (!chain) {
+      continue;
+    }
+
+    if (chain.verificationApi?.type === 'etherscan' && chain.blockExplorerUrl !== supportedChain.blockexplorer) {
+      errors.push(
+        `🔴 ${supportedChain.chainname} (ID: ${supportedChain.chainid}) has a different block explorer URL. Local: ${chain.blockExplorerUrl}, Etherscan: ${supportedChain.blockexplorer}`
```
bdrhn9, line 64:
```suggestion
        `🔴 ${chain.alias} (ID: ${chain.id}) has a different block explorer URL. Local: ${chain.blockExplorerUrl}, Etherscan: ${supportedChain.blockexplorer}`
```

I missed that in my first review. We can make messages more consistent by using chain alias on all warning/error types. 

Similar situation at last check.

PR #615, opened 2025-09-28
Update RPC provider configurations

bdrhn9:
`blastapi` sunsets after October 31, so I removed it from all chain configurations. Similarly, `drpc` started caching the `eth_blockNumber` method too aggressively and this occasionally causes unexpected behavior, so it's better not to use it. Other changes aim to increase the number of providers, and unreliable providers have been either removed or replaced with better alternatives.


hiletmis (approved):
LGTM only a minor issue in changeset

Issue #661, opened 2025-12-01
Add keycard support as alternative to MNEMONIC based signer


hiletmis:
Keycard is a hardware wallet like ledger that utilizes a Java card to perform signing actions. For more info see [keycard.tech](https://keycard.tech)

It is an open-source hardware wallet project backed by [Status](https://get.keycard.tech/pages/keycard). Although Status provides utilities for keycard, a monorepo is created called [hw-wallet](https://github.com/api3dao/hw-wallet) to provide more customizable support across Api3 repositories. 

Monorepo consists of following packages:
* `keycard-manager`
* `keycard-hardhat-provider`

**keycard-manager:** This is the core package that communicates with keycard. Exports basic methods that can allow the user to drive a keycard. The main objective of the package is to wrap keycard with [ethers wallet](https://github.com/api3dao/hw-wallet/blob/main/packages/keycard-manager/src/application/keycard-wallet.ts) class so that implementing keycard can be standardized. Also extends [ethers signer](https://github.com/api3dao/hw-wallet/blob/main/packages/keycard-manager/src/application/keycard-signer.ts) class for various applications.

**keycard-hardhat-provider:** This is a hardhat plugin that based on [**hardhat-ledger**](https://github.com/NomicFoundation/hardhat/tree/v2/packages/hardhat-ledger) (hardhat v2). The plugin extends hardhat config to create a custom provider. In this case it is a keycard signer.

Implementing **keycard-hardhat-provider** will create an alternative to `MNEMONIC` based signer.

PR #662, opened 2025-12-01
Add keycard support

hiletmis:
Closes [**Add keycard support as alternative to MNEMONIC based signer**](https://github.com/api3dao/contracts/issues/661)

This PR adds support for keycard by implementing **keycard-hardhat-provider** plugin. 

* `accounts` object is changed to optional to allow `keycardAccount` to override the main signer.
* `KEYCARD_ACCOUNT` is a wallet address, once it is placed in .env file hardhat config will expect a keycard wallet with specified account. If there is no wallet address `keycardAccount` will not be configured.

**Additional changes:**
A transaction confirmation is [added](https://github.com/api3dao/contracts/commit/0db48d5836b89b957de223640bcb5650b3e34e64) to `deployApi3ReaderProxyV1` to avoid transaction overlap while deploying contracts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "hardhat": "^2.26.2",
     "hardhat-deploy": "1.0.2",
     "hardhat-gas-reporter": "^2.3.0",
-    "keycard-hardhat-provider": "^0.1.1",
+    "keycard-hardhat-provider": "^0.1.2",
     "jest": "^30.0.5",
     "prettier": "^3.5.3",
     "prettier-plugin-solidity": "^2.0.0",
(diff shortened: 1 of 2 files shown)


bdrhn9 (approved):
Thank you for this intellectual approach and the clean implementation. I took the time to review the source code for `keycard-hardhat-provider`, and I learned about how the Hardhat provider can be extended as a side effect of the import.

I tested the following configurations, and they worked as expected:
* Filling `KEYCARD_ACCOUNT` with the correct address.
* Filling `KEYCARD_ACCOUNT` with an incorrect address.
* Filling only the `MNEMONIC`.

The only thing that could be improved is the handling of EIP-55 checksums. I encountered [this](https://github.com/api3dao/hw-wallet/blob/2e8ba85ef1264375a88acf77800feccf20da4a7a/packages/keycard-hardhat-provider/src/provider.ts#L242) error when providing an address that satisfies checksum encoding (e.g., `0x46C6Ef8f5C3f13Ef35d94C4E7E0287b0cdbFCfa2`). It appears the current implementation throws an error because it strictly [expects](https://github.com/api3dao/hw-wallet/blob/2e8ba85ef1264375a88acf77800feccf20da4a7a/packages/keycard-hardhat-provider/src/provider.ts#L237) a lowercase address.

PR #670, opened 2025-12-12
Add Tokos on Somnia

### LINKED ISSUE #669: Add Tokos on Somnia
Add Tokos on Somnia

https://github.com/api3dao/dapp-management/issues/145

aquarat:
Closes #669

### DIFF
--- a/.changeset/light-baths-attack.md
+++ b/.changeset/light-baths-attack.md
@@ -1,0 +1,5 @@
+---
+'@api3/contracts': patch
+---
+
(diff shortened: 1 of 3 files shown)


Siegrift:
See [this](https://github.com/api3dao/dapp-management/issues/145#issuecomment-3275703190). Tokos was at the time of integration still using the old branding and we've used the alias "arenas" for their proxies. The best way here is to do what the comment there says and update the details, but preserve the alias.

PR #776, opened 2026-05-08
Update providers for unichain-sepolia-testnet

bdrhn9:
The current default provider (`https://sepolia.unichain.org`) occasionally causes discrepancies when sending transactions, submitted txes intermittently fail to get mined. The `publicnode` endpoint (`https://unichain-sepolia-rpc.publicnode.com`) has been more reliable in this respect, so promoting it to `default`.


serhhan (approved):
LGTM!

PR #787, opened 2026-07-03
Update providers

bdrhn9:
This PR updates RPC providers for several networks to improve reliability and performance.

## Changes

- **base**: Made `publicnode` (`https://base-rpc.publicnode.com`) the new `default` provider. The previous default (`https://mainnet.base.org`) is kept as a fallback under the `public` alias.
- **core**: Removed `icecreamswap` because it started to not respond.
- **katana**: Removed `conduit` because it returns a gas price significantly higher than usual (around 10000x).
- **moonbeam**: Removed `unitedbloc` because it mostly fails to respond to requests and fails to sync with the head of the chain.
- **sei-testnet**: Removed the previous `default` provider (`https://sei-testnet-public.nodies.app`) because it started to not respond, and promoted the `public` provider (`https://evm-rpc-testnet.sei-apis.com`) to `default`.

### DIFF
--- a/.changeset/spicy-donuts-jump.md
+++ b/.changeset/spicy-donuts-jump.md
@@ -1,0 +1,11 @@
+---
+'@api3/contracts': patch
+---
+
+Update RPC provider configurations:
+
+- Make publicnode default and keep previous default as public for base
+- Remove icecreamswap from core
+- Remove conduit from katana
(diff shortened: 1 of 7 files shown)


bbenligiray (approved):
lgtm

PR #801, opened 2026-07-24
Csur-filecoin

hiletmis:
This PR adds support for following chains:
- filecoin
- filecoin-testnet

### ⚠️ Deployment notes

Contracts were **deployed non-deterministically**, so their addresses **do not match** the shared CREATE2 addresses used on every other chain. This is unavoidable on Filecoin:

- **Deterministic (CREATE2) deployment is impossible on Filecoin — mainnet and testnet alike.** The Arachnid deterministic-deployment proxy (`0x4e59b448…`) is not present and cannot be installed: its keyless presigned deployment transaction is hardcoded to a `100000` gas limit (Nick's method), which is far below what FEVM requires to deploy the factory, and the limit can't be raised without invalidating the signature. Both Filecoin mainnet and testnet therefore use `deploy:undeterministic`.
- `check:deployment-addresses` does not require cross-chain address equality, so this passes CI.

**Reproducibility caveat (testnet only):** the calibration RPC returns Ethereum tx hashes that ethers v5 rejects with `Transaction hash mismatch`, which aborts hardhat-deploy. Deployment required a one-off local preload shim (`filecoin-hash-shim.cjs`) that makes ethers trust the node-reported hash. This is a calibration-RPC quirk only — Filecoin **mainnet** does not need it.


bdrhn9 (approved):
Lgtm 👍🏼

PR #807, opened 2026-07-28
Skip undeterministic deployment verification

### LINKED ISSUE #461: Only allow contracts to be deployed non-deterministically for chains with exclusive permission
Similar to https://github.com/api3dao/contracts/blob/0099d256ab6205c8ebbaa0542c5cf6cb8d4c5f5c/scripts/constants.ts#L14 we can keep an array of chain aliases such as https://github.com/api3dao/contracts/issues/388 and verify-deployments would fail if the contract is deployed non-deterministically for a chain that isn't listed there.

hiletmis:
Closes https://github.com/api3dao/contracts/issues/461

## Problem

`verify-deployments` fails on both Filecoin chains:

```
Error: filecoin GnosisSafeWithoutProxy creation tx could not be fetched
Error: filecoin-testnet GnosisSafeWithoutProxy creation tx could not be fetched
```

All contracts on these two chains are deployed undeterministically, so they are verified
through their creation transactions. Filecoin's Ethereum JSON-RPC implementation only
indexes recent messages, which means `eth_getTransactionByHash` returns `null` for the
deployment transactions and verification cannot proceed.

This is a property of the chains' RPC retention rather than a problem with the
deployments themselves.

## Solution

Add `skippedChainAliasesInUndeterministicDeploymentVerification` to `scripts/constants.ts`,
following the existing skip-list convention in that file. For the listed chains,
`verify-deployments` confirms that contract code exists at the deployment address instead
of fetching and comparing the creation transaction.


scripts/verify-deployments.ts
```diff
@@ -172,6 +176,14 @@ async function verifyDeployments(network: string) {
...
         throw new Error(`${network} ${contractName} (deterministic) contract code does not exist`);
       }
+    } else if (skippedChainAliasesInUndeterministicDeploymentVerification.includes(network)) {
+      const goFetchContractCode = await go(async () => provider.getCode(deployment.address), goAsyncOptions);
+      if (!goFetchContractCode.success || !goFetchContractCode.data) {
+        throw new Error(`${network} ${contractName} (undeterministic) contract code could not be fetched`);
+      }
+      if (goFetchContractCode.data === '0x') {
+        throw new Error(`${network} ${contractName} (undeterministic) contract code does not exist`);
+      }
```
bdrhn9: The `else if` body is identical to the `if` body, except the two error strings use `undeterministic` instead of `deterministic`. I'd suggest the following implementation to reduce duplication:

```ts
const deployedDeterministically = deployment.address === expectedDeterministicDeploymentAddress;
if (deployedDeterministically || skippedChainAliasesInUndeterministicDeploymentVerification.includes(network)) {
  const deploymentType = deployedDeterministically ? 'deterministic' : 'undeterministic';
  const goFetchContractCode = await go(async () => provider.getCode(deployment.address), goAsyncOptions);
  if (!goFetchContractCode.success || !goFetchContractCode.data) {
    throw new Error(`${network} ${contractName} (${deploymentType}) contract code could not be fetched`);
  }
  if (goFetchContractCode.data === '0x') {
    throw new Error(`${network} ${contractName} (${deploymentType}) contract code does not exist`);
  }
} else {
```

bdrhn9 (commented):
I don't think this PR addresses issue #461. Issue #461 asks for the opposite: maintaining an allowlist of chains that were deployed non-deterministically. This is because non-deterministic deployments are normally validated using [this](https://github.com/api3dao/contracts/blob/4eecd492831c125fdbe0eafde1516f18c6daee69/scripts/verify-deployments.ts#L175) `else` condition, and CI passes. For example, ronin, mantle-sepolia-testnet, and taiko have non-deterministic deployments for the OwnableCallForwarder, AccessControlRegistry, and Api3ServerV1 contracts.

For this PR, it's probably better to keep the behavior as-is but have it only say "Skip `verify-deployments` on Filecoin, whose RPC does not index historical transactions". We should then create a follow-up PR, maybe with a function written specifically for this chain, to verify the contract code, probably by calling `eth_getCode` and checking the immutable parts of the code. Otherwise, skipping verification completely isn't acceptable, imo.

bdrhn9 (approved):
Created [this](https://github.com/api3dao/contracts/issues/813) issue as a follow-up. This part looks good to me.

PR #817, opened 2026-08-03
Compare deployed bytecode where the creation tx cannot be fetched

### LINKED ISSUE #813: Verify deployed bytecode for chains in `skippedChainAliasesInUndeterministicDeploymentVerification`
Follow-up to #807. For the chains it skips, `verify-deployments` only checks that `eth_getCode` returns something other than `0x`.

The code is already being fetched, so compare it against `artifact.deployedBytecode`, splitting off the metadata hash the way the creation transaction branch does.

The one catch is immutables. Their values are written into the deployed bytecode at deployment while the artifact has zeros in those slots, so both sides need masking first, using `immutableReferences` from `artifacts.getBuildInfoSync()`. Four of the seven contracts fail an unmasked comparison.

This still will not cover constructor arguments or constructor logic, since neither appears in runtime bytecode.

bdrhn9:
Closes #813

On the chains whose RPCs do not serve historical transactions, `verify-deployments` only checked that the deployment address holds some code, which passes for any address holding any contract. It now compares that code against the local compilation output. 

Immutables are masked on both sides as the issue describes, but the byte ranges come from `deployments.getExtendedArtifact()` rather than `artifacts.getBuildInfoSync()`, which returns the same `immutableReferences` without a second way of reaching the compilation output. Constructor arguments stay unverified on these chains, as the issue notes.

`skippedChainAliasesInUndeterministicDeploymentVerification` is renamed to `chainAliasesWithoutHistoricalTransactionIndexing`, since being on the list now selects a different check rather than skipping one.

`NETWORK=filecoin` and `NETWORK=filecoin-testnet` pass, as does `ronin`, which still takes the untouched creation tx path.


hiletmis (approved):
LGTM 👍 

Thanks for the improvements

PR #834, opened 2026-09-08
Cret-2026-09-07

hiletmis:
Related to [**Cret-2026-09-07**](https://github.com/api3dao/data-feeds/issues/4501)

This PR removes support for following chains:

* filecoin
* filecoin-testnet


scripts/constants.ts
```diff
@@ -15,7 +15,7 @@ export const skippedChainAliasesInOevAuctionHouseNativeCurrencyRateValidation: s
 
 // The creation tx of an undeterministic deployment cannot be fetched on these chains, so its deployed bytecode is
 // compared with the local compilation output instead
-export const chainAliasesWithoutHistoricalTransactionIndexing = ['filecoin', 'filecoin-testnet'];
+export const chainAliasesWithoutHistoricalTransactionIndexing = [];
```
bdrhn9:
```suggestion
export const chainAliasesWithoutHistoricalTransactionIndexing: string[] = [];
```

`chainAliasesWithoutHistoricalTransactionIndexing` is empty now, and with no type annotation TypeScript infers `never[]` rather than `string[]`. That turns the `.includes(network)` call in `verify-deployments.ts` into a type error:

```sh
scripts/verify-deployments.ts(144,91): error TS2345:
  Argument of type 'string' is not assignable to parameter of type 'never'.
````

Nothing actually fails because of it. `tsconfig.build.json` only covers `bin`, `src` and `typechain-types`, and `hardhat` run is transpile-only unless you pass `--typecheck`, so `pnpm verify-deployments` still runs fine. It only shows up in the editor and in `tsc -p tsconfig.json`.

bdrhn9 (approved):
Lgtm but it's worth to address the suggestion.

